PR #10, opened 2016-10-15
Improve error handling with more specific exceptions and require.

hollinwilkins:
Got rid of all "throw new Error" and replaced with IllegalArgumentException or a require statement.


hollinwilkins:
Will reopen after improving error handling for managed resources.

mleap-runtime/src/main/scala/ml/combust/mleap/runtime/bundle/ops/feature/VectorAssemblerOp.scala
```diff
@@ -29,12 +29,12 @@ object VectorAssemblerOp extends OpNode[VectorAssembler, VectorAssemblerModel] {
...
 
   override def shape(node: VectorAssembler): Shape = {
-    val s = Shape()
     var i = 0
-    for(input <- node.inputCols) {
-      s.withInput(input, s"input$i")
-       i = i + 1
-    }
-    s.withStandardOutput(node.outputCol)
+    node.inputCols.foldLeft(Shape()) {
```
prianna: Hey baby, I like the way your `Shape` folds.

PR #11, opened 2016-10-15
Add build info to MLeap and Bundle.ML.

hollinwilkins:
Add in automatically-generated BuildInfo class with version, git SHA and Scala Version.


hollinwilkins:
Will reopen after restructuring the project.

Issue #9, opened 2016-10-14
Implement MLeap UDFs up to 5 input fields.


hollinwilkins:

hollinwilkins:
Pull request in at #16

hollinwilkins:
Merged with pr #16

PR #16, opened 2016-10-21
Add in user defined functions up to 5 arguments

hollinwilkins:
This is part of a larger effort to be able to execute on Spark easier. This will also simplify how we write transformers eventually.

User defined functions use implicits and reflection to allow users to write Scala functions that can operate on data within a LeapFrame.

This is for ticket #9

### DIFF
--- a/mleap-runtime/src/main/scala/ml/combust/mleap/runtime/LeapFrame.scala
+++ b/mleap-runtime/src/main/scala/ml/combust/mleap/runtime/LeapFrame.scala
@@ -118,37 +118,6 @@ trait LeapFrame[LF <: LeapFrame[LF]] extends TransformBuilder[LF] with Serializa
     }
   }
 
-  /** Try to add multiple fields to the LeapFrame.
-    *
-    * Returns a Failure if trying to add any existing fields.
-    *
-    * @param field first field to add
-    * @param fields fields to add
-    * @param f function for calculating new field values
-    * @return try new LeapFrame with new fields
-    */
-  def withFields(field: StructField, fields: StructField *)
-                (f: (Row) => Row): Try[LF] = {
-    withFields(field +: fields)(f)
-  }
-
-  /** Try to add multiple fields to the LeapFrame.
-    *
-    * Returns a Failure if trying to add any existing fields.
-    *
-    * @param fields fields to add
-    * @param f function for calculating new field values
-    * @return try new LeapFrame with new fields
-    */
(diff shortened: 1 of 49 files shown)


hollinwilkins:
Take a look at the https://github.com/combust-ml/mleap/compare/feature/udfs...feature/spark-integration branch to see a larger use for this work. In addition to making Spark integration very simple, it will also allow for easier code generation when we start doing that because there is only one entry point for transformers to build there transformation (TransformBuilder#withInput).

mleap-runtime/src/main/scala/ml/combust/mleap/runtime/LeapFrame.scala
```diff
@@ -50,63 +52,69 @@ trait LeapFrame[LF <: LeapFrame[LF]] extends TransformBuilder[LF] with Serializa
...
-  def withFields(fields: Seq[StructField])
-                (f: (Row) => Row): Try[LF] = {
-    schema.withFields(fields).map {
-      schema2 =>
-        val dataset2 = dataset.withValues(f)
-        withSchemaAndDataset(schema2, dataset2)
+  def withField(name: String, selectors: Selector *)
+               (udf: UserDefinedFunction): Try[LF] = {
+    var i = 0
+    selectors.foldLeft(Try(Seq[RowSelector]())) {
```
prianna: Could just make this a tuple of (trss, acc)

prianna:
```scala
selectors.foldLeft((Try(Seq[RowSelector]()), 0)) {
      case ((trss, acc), s) =>
        (createRowSelector(s, udf.inputs(acc)).flatMap {
          rs => trss.map(trs => rs +: trs)
        }, acc + 1)
    }._1.flatMap {
      rowSelectors =>
        schema.withField(name, udf.returnType).map {
          schema2 =>
            val dataset2 = dataset.withValue(rowSelectors: _*)(udf)
            withSchemaAndDataset(schema2, dataset2)
        }
    }
```

hollinwilkins: This will cause allocations of unneeded tuples, which I want to avoid when possible. Otherwise, I am all for completely functional programming :)

hollinwilkins:
Merging this in to continue down the path of Spark integration and custom types.

PR #19, opened 2016-10-26
Spark Integration and Custom Types

hollinwilkins:
This branch allows for MLeap transformers to execute on Spark DataFrames
This branch allows for the use of custom types in a LeapFrame as well as in the MLeap transformers.

Further changes:
1. Load the registry from reference.conf files instead of manually building them
2. Load custom types for both Bundle.ML and MLeap Runtime from reference.conf files


prianna:
Reviewing today

Issue #58, opened 2016-12-25
Add imputer transformer to MLeap


hollinwilkins:
We should add imputer to MLeap based on the Spark transformer.

1. MLeap core model
2. MLeap runtime transformer
3. MLeap transformer NodeOp
4. Spark transformer NodeOp

PR #98, opened 2017-01-04
Feature/imputer

prianna:
Adds #58. Still some issues with the travis build, need to troubleshoot before merge. Will add tests in a separate branch.


hollinwilkins:
Some things that still need to be done:
1. Add Ops to the reference.conf files
2. Make sure the Spark parity is working for both null and non-null values

PR #124, opened 2017-02-10
Add Spark RegexTokenizer as FeatureTransformer

MeiSign:
I have integrated the RegexTokenizer of Spark as Feature Transformer according to the documentation because I use it in my pipelines and would like to use mleap for pipeline execution.
A merge would be very much appreciated, hope I forgot nothing...


hollinwilkins:
Hey,

Just glanced briefly and it looks great. Will merge this in sometime today.

Thank you for the contribution!

PR #143, opened 2017-03-03
conflict resolution for scalatest between mleap-core and spark-mllib-local

### LINKED ISSUE #142: Incompabilities between scalatest version for mleap-core(test scope) and spark-mllib-local(compile scope)
Running the SBT build shows the following warning 

[warn] There may be incompatibilities among your library dependencies.
[warn] Here are some of the libraries that were evicted:
[warn] 	* org.scalatest:scalatest_2.10:2.2.6 -> 3.0.0
[warn] Run 'evicted' to see detailed eviction warnings

In SBT, this causes no problems however, when Intellij runs its version of the SBT build, some tests give the following error, for example:

Error:(19, 13) not enough arguments for macro method assert: (implicit prettifier: org.scalactic.Prettifier, implicit pos: org.scalactic.source.Position)org.scalatest.Assertion.
Unspecified value parameters prettifier, pos.
      assert(km(Vectors.dense(Array(111.0, 20.0, 56.0))) == 2)

This is caused by the fact that while mleap-core brings in scalatest dependency version 3.0.0 in test scope, spark-mllib-local actually brings in scalatest dependency version 2.2.6 in compile scope.

ancasarb:
Fixes #142 by including scalatest 3.0.0 at compile time in mleap-core because spark-mllib-local requires the scalatest at compile time due to its dependency on spark-tags. However, the only scalatest bit that's used in spark-tags is "import org.scalatest.TagAnnotation;" so this works the same in versions 2.2.6 and 3.0.0.

Alternatively, in Dependencies.scala, we could exclude the scalatest version brought by spark-mllib-local, if not having scalatest as a compile dependency is more preferable 

```
val sparkMllibLocal = "org.apache.spark" %% "spark-mllib-local" % sparkVersion excludeAll(ExclusionRule(organization = "org.scalatest"))
```

and the tests and build still pass.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -51,7 +51,7 @@ object Dependencies {
 
   val base = l ++= Seq()
 
-  val core = l ++= Seq(sparkMllibLocal, jTransform, Test.scalaTest)
+  val core = l ++= Seq(sparkMllibLocal, jTransform, scalaTest)
 
   def runtime(scalaVersion: SettingKey[String]) = l ++= (Seq(Test.scalaTest) ++ scalaReflect.modules(scalaVersion.value))
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -13,7 +13,7 @@ object Dependencies {
   val akkaHttpVersion = "10.0.3"
 
   object Compile {
-    val sparkMllibLocal = "org.apache.spark" %% "spark-mllib-local" % sparkVersion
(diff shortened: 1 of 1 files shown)


hollinwilkins:
@ancasarb Hey, thank you for figuring out the root cause of this issue and the PR. Can we go with your other proposed solution of excluding scalatest from the compile dependencies of mllib-local? I would prefer this to having an unnecessary compile dependency.

PR #145, opened 2017-03-06
Update "/transform" path in mleap-serving module to only accept POST method

ancasarb:
Changes to update "/transform" path in MleapResource (mleap-serving module) to only accept POST method as documented in the README.md.

While reading about how mleap-serving module works, I've also added scalatest and akka-http-testkit as testing dependencies to mleap-serving module and tests for MleapService and MleapResource.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,7 @@
 import ml.combust.mleap.MleapProject
 
+enablePlugins(CrossPerProjectPlugin)
+
 lazy val root = MleapProject.root
 lazy val base = MleapProject.baseProject
 lazy val bundleMl = MleapProject.bundleMl
--- a/mleap-serving/src/main/resources/application.conf
+++ b/mleap-serving/src/main/resources/application.conf
@@ -6,5 +6,3 @@ ml.combust.mleap.serving {
     bind-port = ${ml.combust.mleap.serving.http.port}
   }
 }
-
-ml.blah=hey
--- a/mleap-serving/src/main/scala/ml/combust/mleap/serving/MleapResource.scala
+++ b/mleap-serving/src/main/scala/ml/combust/mleap/serving/MleapResource.scala
@@ -45,8 +45,10 @@ class MleapResource(service: MleapService)
             complete(service.unloadModel(UnloadModelRequest()))
           }
         } ~ path("transform") {
-          entity(as[DefaultLeapFrame]) {
-            frame => complete(service.transform(frame))
+          post {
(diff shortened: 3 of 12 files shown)


ancasarb:
I'll take a look into the issue of "unresolved dependency: com.typesafe.akka#akka-http_2.10;10.0.3" causing the Travis CI build to fail shortly.

hollinwilkins:
@ancasarb Thanks for this awesome PR. The issue you are running into is that Akka does not support Scala 2.10 anymore, so you will have to disable cross-compilation for MLeap-serving. This can be done by making some settings for it in Common.settings.

ancasarb:
@hollinwilkins I'm still relatively new to sbt so am not too sure how to disable cross-compilation for MLeap-serving. I see

```
lazy val defaultMleapServingSettings = defaultMleapSettings ++ Seq(crossScalaVersions := Seq("2.11.8"))
```

in Common.scala already but I'm guessing that's not enough?

Thanks for your help!

ancasarb:
@hollinwilkins I got it to work by using sbt plugin sbt-doge to aggregate respecting crossScalaVersions at each module, could you please take a look?

Also, since mleap-serving module is now included in the aggregatedProjects, it will be published to nexus as part of the CI build, is that alright? Alternatively, I could remove it from aggregatedProjects and update the sbt command in travis.sh to run the mleap-serving tests as well, what do you think?

Thank you!

ancasarb:
It seems that mleap-serving module is using the default akka.event.Logging$DefaultLogger logger that logs to STDOUT and is registered by default and not a slf4j logger. I've silenced the error logging for testing in the application.conf included in src/test/resources.

hollinwilkins:
@ancasarb Hey, sorry but I had to revert this change. It is breaking the publish to maven step. Not sure why, but I suspect it has something to do with sbt-doge. We'll have to figure out how to fix that issue.

ancasarb:
Really sorry about that. As an alternative to sbt-doge plugin, I could remove mleap-serving from the aggregatedProjects list (as it was before) and add the mleap-serving/test to the sbt command in travis.sh, what do you think?

hollinwilkins:
I like that idea. I think for the most part people will use mleap-serving entirely separate from the rest of MLeap. Once we migrate to 2.11/2.12 and end support for 2.10 we can add it back in to aggregate projects.

ancasarb:
Added https://github.com/combust/mleap/pull/147 for the same changes without sbt-doge plugin, thank you!

PR #156, opened 2017-03-16
Update equality of tensors to also check same values

ancasarb:
I think that, if my understanding is correct, two tensors with same dimension but different elements should not be equal, i.e.

val tensor1 = Tensor.denseVector(Array(20.0, 11.0, 5.0))
val tensor2 = Tensor.denseVector(Array(20.0, 10.0, 5.0))

should not be considered equal. 

I've updated the equals method for both dense and sparse tensors to reflect this and added some spec tests for the equals method.

### DIFF
--- a/mleap-spark/src/test/scala/org/apache/spark/ml/parity/clustering/LDAParitySpec.scala
+++ b/mleap-spark/src/test/scala/org/apache/spark/ml/parity/clustering/LDAParitySpec.scala
@@ -5,10 +5,13 @@ import org.apache.spark.ml.clustering.LDA
 import org.apache.spark.ml.feature.{CountVectorizer, StopWordsRemover, Tokenizer}
 import org.apache.spark.ml.parity.SparkParityBase
 import org.apache.spark.sql.DataFrame
+import org.scalatest.Ignore
 
 /**
   * Created by mageswarand on 14/3/17.
   */
+@Ignore
+// to re-enable as part of https://github.com/combust/mleap/issues/169
 class LDAParitySpec extends SparkParityBase {
   override val dataset: DataFrame = textDataset.select("text")
 
--- a/mleap-tensor/src/main/scala/ml/combust/mleap/tensor/Tensor.scala
+++ b/mleap-tensor/src/main/scala/ml/combust/mleap/tensor/Tensor.scala
@@ -99,8 +99,8 @@ case class DenseTensor[T](values: Array[T],
           if (obj.values.isEmpty) { true }
           else { false }
         } else {
-          (dimensions == obj.dimensions) ||
-            (dimensions.head == -1 || obj.dimensions.head == -1 && dimensions.tail == obj.dimensions.tail) &&
+          ((dimensions == obj.dimensions) ||
+            ((dimensions.head == -1 || obj.dimensions.head == -1) && dimensions.tail == obj.dimensions.tail)) &&
               (values sameElements obj.asInstanceOf[DenseTensor[T]].values)
         }
       } else { false }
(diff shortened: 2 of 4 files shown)


ancasarb:
@hollinwilkins seems that this change is causing LDAParitySpec test to fail

`
[info] LDAParitySpec:
[info] - has parity between Spark/MLeap *** FAILED ***
[info]   scala.this.Predef.refArrayOps[org.apache.spark.sql.Row](sparkDataset).sameElements[org.apache.spark.sql.Row](scala.this.Predef.wrapRefArray[org.apache.spark.sql.Row](mleapDataset)) was false (SparkParityBase.scala:120)
`

I wonder if my assumption that values should also be checked is not correct.

hollinwilkins:
@ancasarb I thought we were checking element equality before. It's difficult to see what the parenthesis change, but they obviously do change the equality test somehow. Can you describe what was wrong before and what those parens do?

ancasarb:
@hollinwilkins I was seeing that with two tensors (for example the ones below) with the same dimensions but different elements, this was still returning true, instead of false. 

`   
 it("should return false for dense tensors with same dimension but different elements") {
      val tensor1 = Tensor.denseVector(Array(20.0, 11.0, 5.0))
      val tensor2 = Tensor.denseVector(Array(20.0, 10.0, 5.0))
      assert(tensor1 != tensor2)
    }
`

hollinwilkins:
@ancasarb I see, this probably indicates an issue with LDA then. Will have to investigate further.

ancasarb:
I think it was because the equality was checking (dimensions == obj.dimensions) and since this was included in an OR statement and it was returning true then it didn't actually check the elements.

ancasarb:
@hollinwilkins should I close the PR for now while we investigate?

hollinwilkins:
Let's keep it open. It's correct, just need to figure out why LDA doesn't produce the same values.

hollinwilkins:
@Mageswaran1989 Hey,

When we updated the Tensor equality code to make sure two tensors have the same values in them, we notice that the LDA parity spec breaks. Would you be able to investigate this a bit or help us debug?

Thank you!

ancasarb:
The tensor (dense) that is causing the failure has the following values, which are similar, but not equal

sparkDataset tensor values
 0 = 0.028954998227547032
 1 = 0.8836739663256541
 2 = 0.028936401930665906
 3 = 0.028797922505784255
 4 = 0.029636711010348833

mleapDataset tensor values
 0 = 0.02895479212197769
 1 = 0.8836912061545042
 2 = 0.028919756312963223
 3 = 0.02879775419971172
 4 = 0.029636491210843156

hollinwilkins:
Interesting, I wonder if we can relax the constraints for the LDA parity spec or why the numbers are off by any amount in the first place. I'm unfamiliar with how this transformer works.

Mageswaran1989:
hello...I will get back to you after doing a sanity test on our pipeline sometime today

Mageswaran1989:
The [optimizer used in LDA](https://github.com/combust/mleap/blob/master/mleap-core/src/main/scala/ml/combust/mleap/core/clustering/optimization/LDAOptimizer.scala#L37)  does a random sampling which most likely be the cause for this mismatch.

analysis with python:    
```
import numpy as np
a0 = 0.028954998227547032  
a1 = 0.8836739663256541  
a2 = 0.028936401930665906  
a3 = 0.028797922505784255  
a4 = 0.029636711010348833   
```

```
#mleapDataset tensor values   
b0 = 0.02895479212197769   
b1 = 0.8836912061545042   
b2 = 0.028919756312963223   
b3 = 0.02879775419971172   
b4 = 0.029636491210843156   
```

```
np.set_printoptions(precision=16, suppress=True)
np.array(alist) - np.array(blist)  
```
**o/p**
array([ 0.0000002061055693, -0.0000172398288502,  0.0000166456177027,
        0.0000001683060725,  0.0000002197995057])

We instead validated by checking the sum of both the predictions, which in this case is
`a1 + a2 + a3 + a4 + a0` =>  1.0000000000000002
`b1 + b2 + b3 + b4 + b0` => 1.0

hollinwilkins:
@Mageswaran1989 Thank you for all of the info. We may disable the LDA parity spec for now in favor of getting this core fix of tensor equality into MLeap. I will make a ticket to reenable it before the next release.

@ancasarb Can you comment out the LDA parity spec in this branch for now and I will merge it in. I think this fix is very important as it is a major bug fix.

For reference, here is the ticket to fix LDA parity spec to be more forgiving about small differences in tensor values:
https://github.com/combust/mleap/issues/169

ancasarb:
@hollinwilkins I've updated the branch to ignore the LDA test for now, could you please have a look? thank you!

Issue #135, opened 2017-02-24
empty OneHotEncoder.node in root folder


kaixinyan:
when I'm including oneHotEncoder in machine learning pipeline, fitting data using the pipeline, and then saving the model as a zip file, "oneHot_4cde23ce6de5.node" is an empty folder, all other folders, say "strIdx_0dc5e14fb06e.node", have two files, "model.json" and "node.json" inside. Also, the unzipped file is a folder called "root", there is no "bundle.json". If I remove oneHotEncoder transformer, everything is fine.

The code is attached below. Any help is appreciated.

import org.apache.spark.ml.{Pipeline, PipelineModel}
import org.apache.spark.ml.classification.LogisticRegression
import org.apache.spark.ml.feature.{HashingTF, Tokenizer, OneHotEncoder, VectorAssembler}
import org.apache.spark.ml.linalg.Vector
import org.apache.spark.sql.Row

val training = spark.createDataFrame(Seq(
  (0L, 1, 1.0),
  (1L, 1, 0.0),
  (2L, 0, 1.0),
  (3L, 0, 0.0)
)).toDF("id", "gender", "label")

val oneHotEncoder = new OneHotEncoder()
     .setInputCol("gender")
     .setOutputCol("genderVectored")

val featAssembler = new VectorAssembler()
      .setInputCols(Array("genderVectored"))
      .setOutputCol("features")

val lr = new LogisticRegression()
  .setMaxIter(10)
  .setRegParam(0.001)

val pipeline = new Pipeline()
  .setStages(Array(oneHotEncoder, featAssembler, lr))

val model = pipeline.fit(training)

hollinwilkins:
Because OneHotEncoder operates on meta data in the DataFrame to know how large the output arrays are, you need to provide a transformed DataFrame when saving a pipeline with a one hot encoder. I'll put up an example on the docs for tomorrow. In the meantime, you can have a look here where we provide a transformed data frame to the SparkBundleContext before serializing:

https://github.com/combust/mleap/blob/master/mleap-spark-testkit/src/main/scala/org/apache/spark/ml/parity/SparkParityBase.scala#L109

Note: When you do this, the transform is not actually being performed, you are just generating the schema of the data frame, so it is low cost.

kaixinyan:
thanks Hollin. After adding StringIndexer and using (context) defined below, it now works fine.

implicit val context = SparkBundleContext().withDataset(sparkTransformed)

[...]

PR #159, opened 2017-03-30
IndexToString labels parameter should be optional

pnpritchard:
When I use `IndexToString`, I do not set the `labels` parameter so the MLeap serialization fails. Spark documentation agrees that this is an optional parameter: https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/feature/StringIndexer.scala#L321. I'm thinking there should be a test case for this, but wasn't sure how to do this without creating another `ReverseStringIndexerParitySpec` class.

### DIFF
--- a/mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/ReverseStringIndexerOp.scala
+++ b/mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/ReverseStringIndexerOp.scala
@@ -4,6 +4,7 @@ import ml.combust.bundle.BundleContext
 import ml.combust.bundle.op.{OpModel, OpNode}
 import ml.combust.bundle.dsl._
 import org.apache.spark.ml.bundle.SparkBundleContext
+import org.apache.spark.ml.bundle.util.ParamUtil
 import org.apache.spark.ml.feature.IndexToString
 
 /**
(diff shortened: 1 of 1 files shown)


hollinwilkins:
@pnpritchard Hey Nick,

Spark does indeed allow for optional setting of the labels. If you set it to None, it will use the meta data from the DataFrame to infer the values, similar to what OneHotEncoder does for determining the size of its output vector. MLeap models should always have the labels set, but when serializing the Spark model, we should do this:

1. Check if labels are set, if they are, great, serialize them to the bundle and be done
2. If labels are not set, check to see if the user provided us with a DataFrame for meta data information. If they did not provide the data frame, throw an error, if they did provide a data frame then extract the labels from it

Take a look at OneHotEncoder for an example of how we can use user-supplied data frames to get meta data information like this. 
https://github.com/combust/mleap/blob/master/mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/OneHotEncoderOp.scala#L47-L50

Also take a look at this issue here: https://github.com/combust/mleap/issues/135

pnpritchard:
@hollinwilkins Thanks for the feedback

Just to confirm, the reason labels are necessary in the bundle is for non-Spark runtimes (e.g. MLeap) that do no have an equivalent of DataFrame metadata?

I will go ahead and make the changes you suggested. Apologies for my misunderstanding, I have just been using MLeap for serialization/deserialization and not as a runtime (yet).

hollinwilkins:
@pnpritchard That's right. There is no concept of meta data in a LeapFrame (our lightweight DataFrame used in the MLeap runtime). This is for performance reasons, as calculating meta data can be expensive.

PR #181, opened 2017-04-19
Adding datatype to input/output socket as a first step needed for issue 38, providing schema of the deployed model

ancasarb:
@hollinwilkins @seme0021 

I've given it a try to add a Datatype to the node input/output socket in order to be able to provide the schema of the deployed model, updates to the protobuf messages are in [this](https://github.com/combust/bundle-protobuf/compare/develop...ancasarb:feature/issue38_socket_type?expand=1) branch.

I've used the DataFrame that some of the metadata dependent transformers require to be passed in when serializing the pipeline for setting the types. If the data frame is passed in the spark context, the datatype for the node input / output sockets is derived from it, otherwise the type is set to null. (I will add a test for this in the next commit)

When serializing a model from mleap, all datatypes will be set to null for the moment, but I can make the same changes to the MleapContext if you think this approach works.  

@hollinwilkins @seme0021 I'd appreciate your thoughts or ideas on whether you think this is feasible, thank you!

Sidenotes:
- I've added a method to convert from spark types to bundle ml / mleap types in TypeConverters.scala for all the spark types that I found to be used. 

- I've updated four of the ops (MultiLayerPerceptronClassifierOp, GaussianMixtureOp, MathBinaryOp, MathUnaryOp) in the spark subproject that were using the MleapContext when I think they should be using the SparkBundleContext instead.

### DIFF
--- a/bundle-ml/src/main/scala/ml/combust/bundle/BundleContext.scala
+++ b/bundle-ml/src/main/scala/ml/combust/bundle/BundleContext.scala
@@ -1,6 +1,5 @@
 package ml.combust.bundle
 
-import java.io.File
 import java.nio.file.{FileSystem, Path}
 
 import ml.combust.bundle.serializer.{ConcreteSerializationFormat, SerializationContext, SerializationFormat}
--- a/bundle-ml/src/main/scala/ml/combust/bundle/dsl/Shape.scala
+++ b/bundle-ml/src/main/scala/ml/combust/bundle/dsl/Shape.scala
@@ -175,6 +175,16 @@ case class Shape private (inputs: Seq[Socket],
     name.map(n => withOutput(n, port)).getOrElse(this)
   }
 
+  /** Add an optional typed output socket to the shape.
+    *
+    * @param name name of optional output socket
+    * @param port port of output socket
+    * @return copy of the shape with output socket optionally added
+    */
+  def withOutput(name: Option[String], port: String, dataType: Option[DataType]): Shape = {
+    name.map(n => withOutput(n, port, dataType)).getOrElse(this)
+  }
+
   /** Get the bundle protobuf shape.
     *
     * @return bundle protobuf shape
--- a/bundle-ml/src/main/scala/ml/combust/bundle/dsl/Shape.scala
+++ b/bundle-ml/src/main/scala/ml/combust/bundle/dsl/Shape.scala
@@ -1,5 +1,6 @@
 package ml.combust.bundle.dsl
 
+import ml.bundle.DataType.DataType
 import ml.bundle.Socket.Socket
 
 /** Companion object for holding constant values.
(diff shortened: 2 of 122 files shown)


ancasarb:
The travis build fails because it cannot find revision f6e0e9b2c52b1c8df461e8ed40d9051a63b06715 of bundle-protobuf module at the moment

hollinwilkins:
@ancasarb Hey Anca,

We should definitely discuss this more. An alternative approach is to include the data types on the model object. We could then add a method onto the transformer that allows it to calculate the input/output schema based on an overloaded method for each transformer. This also required less modification to protobuf/MLeap (we already include DataType attributes for some models).

I am not sure which is the best way to go. Including the data types on the node shapes seems very clean to me, but also complicates an otherwise dead simple concept: inputs/outputs. Putting the data types on the models means the schema isn't immediately inferrable and requires extra processing to compute it.

I am leaning slightly towards the original design, the one you implemented, but I would definitely like your thoughts on it.

ancasarb:
@hollinwilkins do you mean doing something similar to this for storing the datatype with the model object (for linear regression for e.g.)

`override def store(model: Model, obj: LinearRegressionModel)
                      (implicit context: BundleContext[SparkBundleContext]): Model = {
      val dataset = context.context.dataset
      model.withAttr("coefficients", Value.vector(obj.coefficients.toArray)).
        withAttr("intercept", Value.double(obj.intercept)).
        withAttr("input_names", Value.string(obj.getFeaturesCol)).
        withAttr("input_types", Value.dataType(fieldType(obj.getFeaturesCol, dataset).get))
    }
`
I found an example of this in the TensorflowTransformer op, I was wondering if there are other transformers that you could point me to? If there are already several transformers doing this, I suppose we wouldn't want to duplicate this info in both node and model objects?

The node object felt like a natural place to put the info regarding the input/output types, as to me, I think of looking at the model object for attributes other than inputs/outputs needed by the op in order to execute while looking at the node object for inputs and outputs. I do agree though that this approach involves more modifications to the protobuf messages, which is more involved than just adding the extra attributes to the model object.

hollinwilkins:
@ancasarb I think the argument for putting this information on the model is that you don't actually have to add it for most models. Like the linear regression model example you gave. We know the coefficients are a Tensor[Double], so therefore we can infer the input/output types are Tensor[Double] and Double. There is only one model I can think of that this doesn't hold true for: the VectorAssembler. Because of the way it works, it's pretty much impossible to tell which inputs are doubles, optional doubles, tensors, etc. So maybe this is a good argument for option 2. We would just modify VectorAssembler to contain the input type information, and then we can infer the information for all the other models.

What do you think?

ancasarb:
@hollinwilkins I see your point, it makes a lot of sense :)  I like the idea of inferring the input /output type information for most of the models. In addition to the VectorAssembler, I think Binarizer model is another example, because as far as I can tell, there's no way to tell if its inputs is a double or a vector, but as you said, there won't be too many such cases.

I'll close this PR for now and give this option 2 a try to implement it, does that sound ok?

hollinwilkins:
@ancasarb Thanks for taking a look at this issue, and yeah, let's try for option 2 for now. Let's try to keep the additions optional for VectorAssembler and Binarizer so that it doesn't mess with serialization too much. Eventually maybe we will require it, but for now I think leaving it optional is fine.

Issue #201, opened 2017-05-02
Spark DecisionTreeModel deserialization issue


pnpritchard:
`SparkNodeWrapper` does not set the `prediction` field in `LeafNode` properly when deserializing bundles. This is a problem because Spark uses the `prediction` field when the classifier/model doesn't have the raw prediction or probability columns. See reproducible case below. I'll submit a PR fix soon.

```scala
val data = sc.parallelize(Seq(
  LabeledPoint(0.0, Vectors.dense(0.0)),
  LabeledPoint(1.0, Vectors.dense(1.0)),
  LabeledPoint(2.0, Vectors.dense(2.0)),
  LabeledPoint(3.0, Vectors.dense(3.0))
)).toDF()

val model = new DecisionTreeClassifier()
  .setRawPredictionCol("")
  .setProbabilityCol("")
  .fit(data)

val file = BundleFile("file:/tmp/model")
model.writeBundle.format(SerializationFormat.Protobuf).save(file)
val model2 = file.loadSparkBundle().get.root

model.transform(data).show()
/*
+-----+--------+----------+
|label|features|prediction|
+-----+--------+----------+
|  0.0|   [0.0]|       0.0|
|  1.0|   [1.0]|       1.0|
|  2.0|   [2.0]|       2.0|
|  3.0|   [3.0]|       3.0|
+-----+--------+----------+
*/

model2.transform(data).show()
/*
+-----+--------+----------+
|label|features|prediction|
+-----+--------+----------+
|  0.0|   [0.0]|       1.0|
|  1.0|   [1.0]|       0.0|
|  2.0|   [2.0]|       0.0|
|  3.0|   [3.0]|       0.0|
+-----+--------+----------+
*/
```

hollinwilkins:
@pnpritchard I believe this is fixed now, closing this issue.

PR #202, opened 2017-05-02
ref #201: fix for setting prediction field

pnpritchard:


hollinwilkins:
@pnpritchard This is great! Thank you for finding this bug. I'm wondering if you can think of any easy tests we can add to illustrate this bug and the fix? Also, have you looked through the MLeap-Core model to see if this will be an issue there as well?

Thank you again for the help

pnpritchard:
@hollinwilkins You're right, the MLeap-Core model has the same issue when using DecisionTreeRegressionModel. I'll commit a fix for that and try working on a test as well. Also, do you use an IDE? I tried adding the sbt-idea plugin so I could use IntelliJ but it didn't work.

hollinwilkins:
@pnpritchard I do use an IDE. Lots of people have trouble with using an IDE for some reason. Can you also post what errors you get when trying to load MLeap into your IDE so we can try to fix it?

Thank you!

hollinwilkins:
Merging this PR in. Actual predictions coming back from decision trees/random forest classifiers are correct because we use probabilities instead of this prediction field directly. We should still have the prediction field be corrected by this PR.

Thanks for the catch!

Issue #164, opened 2017-04-04
Issue in deserializing protobuf models > 64MB


saurabh3949:
Hi,
I was trying to load a big MultiLayer perceptron bundle into Mleap serving. It throws the following error:

[ERROR] [04/03/2017 23:57:48.917] [MleapServing-akka.actor.default-dispatcher-89] [MleapResource] error with request
com.google.protobuf.InvalidProtocolBufferException: Protocol message was too large.  May be malicious.  Use CodedInputStream.setSizeLimit() to increase the size limit.
	at com.google.protobuf.InvalidProtocolBufferException.sizeLimitExceeded(InvalidProtocolBufferException.java:134)
	at com.google.protobuf.CodedInputStream$StreamDecoder.readRawBytesSlowPath(CodedInputStream.java:2768)
	at com.google.protobuf.CodedInputStream$StreamDecoder.readBytes(CodedInputStream.java:2335)
	at ml.bundle.Tensor.Tensor.mergeFrom(Tensor.scala:92)
	at ml.bundle.Tensor.Tensor.mergeFrom(Tensor.scala:11)
	at com.trueaccord.scalapb.LiteParser$.readMessage(LiteParser.scala:15)
	at ml.bundle.Value.Value.mergeFrom(Value.scala:94)
	at ml.bundle.Value.Value.mergeFrom(Value.scala:13)
	at com.trueaccord.scalapb.LiteParser$.readMessage(LiteParser.scala:15)
	at ml.bundle.Attribute.Attribute.mergeFrom(Attribute.scala:54)
	at ml.bundle.Attribute.Attribute.mergeFrom(Attribute.scala:11)
	at com.trueaccord.scalapb.LiteParser$.readMessage(LiteParser.scala:15)
	at ml.bundle.AttributeList.AttributeList$AttributesEntry.mergeFrom(AttributeList.scala:131)
	at ml.bundle.AttributeList.AttributeList$AttributesEntry.mergeFrom(AttributeList.scala:87)
	at com.trueaccord.scalapb.LiteParser$.readMessage(LiteParser.scala:15)
	at ml.bundle.AttributeList.AttributeList.mergeFrom(AttributeList.scala:44)
	at ml.bundle.AttributeList.AttributeList.mergeFrom(AttributeList.scala:11)
	at com.trueaccord.scalapb.LiteParser$.parseFrom(LiteParser.scala:9)
[...]

PR #216, opened 2017-05-25
Increasing protobuf read size

Mageswaran1989:
Fix for https://github.com/combust/mleap/issues/164


Mageswaran1989:
@hollinwilkins here is the fix for #164 
along with this also adding changes for my previous MR. My bad I had to change a class name and has an update for the transformer to work with Spark serialisation.

If you found this needs to be in different MR, let me know.

hollinwilkins:
Looks good, will merge in after tests pass.

Issue #117, opened 2017-02-07
Nice Java Interface


hollinwilkins:
Currently working with MLeap from Java can be a pain. Let's make the interface nicer.

hollinwilkins:
This will most likely be implemented as a combination of the following:

1. Making sure case classes have constructors that are compatible with Java
2. Builders that are Java-friendly
3. Java-friendly wrapper classes when all else fails

bipptech:
`val bundle = (for (bundleFile <- managed(BundleFile("jar:file:/tmp/mleap-model.zip"))) yield {
      bundleFile.loadMleapBundle().get
    }).opt.get

    val schema = StructType(StructField("id", IntegerType()),
      StructField("description", StringType()), StructField("label", StringType())).get
    val data = LocalDataset(Row(1, "random text"))
    val frame = LeapFrame(schema, data)

    val mleapPipeline = bundle.root
    val frame2 = mleapPipeline.transform(frame).get
    frame2.printSchema()
    frame2.show()`

I am writing an evaluator/predictor in java which will be deployed on tomcat. Could you help me convert above  scala code in java?

hollinwilkins:
@bipptech Executing Scala code from Java can be difficult for a variety of reasons. MLeap's biggest hurdle is all of the implicits it depends on. Let me flip this question on you. How would you like to see that code written in Java? We can work from there to design the Java interface to MLeap.

bipptech:
First of all, i would like to thank you for your great service to ML community. Earlier I used to use spark in client mode for predictions and now I use MLeap. Its performance is beyond comparison. 

I wrote a scala object 
```
object MLeapPredictor {
  private val bundle = (for (bundleFile <- managed(BundleFile("jar:file:/tmp/mleap-model.zip"))) yield {
    bundleFile.loadMleapBundle().get
  }).opt.get

  private val schema = StructType(StructField("id", IntegerType()),
    StructField("description", StringType())).get

  def predict(description: String) = {
    val data = LocalDataset(Row(1, description))
    val frame = LeapFrame(schema, data)
    val mleapPipeline = bundle.root
    val frame2 = mleapPipeline.transform(frame).get.select("predictedLabel").get
    val data2 = frame2.dataset
    data2(0).getString(0)
  }
} 
```
and called  ``` MLeapPredictor.predict("<description>") ```  from a java class. Both these classes (scala and java ones) are bundled in a war and deployed in tomcat. It works perfectly fine.

[...]

PR #227, opened 2017-06-11
Adding ability to read mleap frame from json in Java

ancasarb:
#117 adding support so that DefaultFrameReader can be used from Java as well.

### DIFF
--- a/mleap-runtime/src/main/java/ml/combust/mleap/runtime/javadsl/BundleBuilder.java
+++ b/mleap-runtime/src/main/java/ml/combust/mleap/runtime/javadsl/BundleBuilder.java
@@ -10,7 +10,7 @@ import java.io.File;
  * Created by hollinwilkins on 4/21/17.
  */
 public class BundleBuilder {
-    private BundleBuilderSupport support = new BundleBuilderSupport();
+    private final BundleBuilderSupport support = new BundleBuilderSupport();
 
     public BundleBuilder() { }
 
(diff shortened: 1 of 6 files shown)


ancasarb:
Yep, you're right. Will update it shortly.

ancasarb:
@hollinwilkins I think I will actually close the PR without any changes required. I found that it's quite easy to create the FrameReader and FrameWriters from Java.

The code for the class above would end up looking like 

```
package ml.combust.mleap.runtime.javadsl;

import ml.combust.mleap.runtime.DefaultLeapFrame;
import ml.combust.mleap.runtime.serialization.FrameReader;
import ml.combust.mleap.runtime.serialization.FrameWriter;

public class FrameSerializationSupport {

    public static FrameReader getReader(String format) {
        switch(format) {
            case "application/json":
                return new ml.combust.mleap.json.DefaultFrameReader();
            case "application/binary": return new ml.combust.mleap.binary.DefaultFrameReader();
            case "application/avro": return new ml.combust.mleap.avro.DefaultFrameReader();
            default: throw new IllegalArgumentException("invalid format for leap frame serialization: " + format);
        }
    }

    public static FrameWriter getWriter(DefaultLeapFrame frame, String format) {
        switch(format) {
            case "application/json": return new ml.combust.mleap.json.DefaultFrameWriter<>(frame);
            case "application/binary": return new ml.combust.mleap.binary.DefaultFrameWriter<>(frame);
            case "application/avro": return new ml.combust.mleap.avro.DefaultFrameWriter<>(frame);
            default: throw new IllegalArgumentException("invalid format for leap frame serialization: " + format);
        }
    }
}
```

and I don't think this class is  needed. Let me know if you think otherwise and I'll reopen the PR.

PR #229, opened 2017-06-13
Add in time created to bundle.json

jacobfoard:
One of the use cases we've had  is keeping our models up to date on a nightly basis and having alerts for when they're X hours old. So I added a field to the bundle case class and have it auto populate when the bundle is being created. I also couldn't get this to build with the `0.12.head` version of tensorflow so I updated to the latest stable version and the unit tests passed. I can revert that if need be for the time stamp.

### DIFF
--- a/bundle-ml/src/main/scala/ml/combust/bundle/dsl/Bundle.scala
+++ b/bundle-ml/src/main/scala/ml/combust/bundle/dsl/Bundle.scala
@@ -1,7 +1,9 @@
 package ml.combust.bundle.dsl
 
 import java.nio.file.{FileSystem, Path}
+import java.time
 import java.util.UUID
+import java.time.LocalDateTime
 
 import ml.combust.mleap.BuildValues
 import ml.combust.bundle.{BundleContext, BundleRegistry}
@@ -104,7 +106,8 @@ object Bundle {
     apply(BundleInfo(uid = UUID.randomUUID(),
       name = name,
       format = format,
-      version = Bundle.version), root)
+      version = Bundle.version,
+      LocalDateTime.now()), root)
   }
 }
 
@@ -114,11 +117,13 @@ object Bundle {
   * @param name name of the bundle
   * @param format serialization format of the [[Bundle]]
   * @param version Bundle.ML version used for serializing
+  * @param timeCreated LocalDateTime when the model was created
   */
 case class BundleInfo(uid: UUID,
                       name: String,
                       format: SerializationFormat,
-                      version: String)
+                      version: String,
+                      timeCreated: LocalDateTime)
 
 /** Root object for serializing Bundle.ML pipelines and graphs.
   *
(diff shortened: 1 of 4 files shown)


hollinwilkins:
@jacobfoard Thank you for the PR. Will get a chance to look closer at it after DataWorks summit tomorrow. Thanks for updating TensorFlow as well, did you get a chance to test it out with the latest 1.1 version?

jacobfoard:
I've looked into it some more and it doesn't pass the tests because the TensorFlowTransformer doesn't comply to the Transfor abstract with the new getFields requirement. I made an attempt to fix it, but I'm less knowledgeable in that area. I'll post the getFields implementation I did and get some feedback and I'll add that in. I believe that is the only thing stopping the tests from passing.

Sent from my iPhone
________________________________

hollinwilkins:
@jacobfoard Cool, take a swing at it, if it proves to be too tricky at this time, we can return a failure there for now until we have a good solution for it. Anca and I are going to be reviewing the typing system that we have for MLeap in the next couple of weeks anyways.

bundle-ml/src/main/scala/ml/combust/bundle/dsl/Bundle.scala
```diff
@@ -104,7 +105,8 @@ object Bundle {
     apply(BundleInfo(uid = UUID.randomUUID(),
       name = name,
       format = format,
-      version = Bundle.version), root)
+      version = Bundle.version,
+      LocalDateTime.now().toString), root)
```
hollinwilkins: Does this return iso8601 formatted dates? We should ensure this is the format we are using.

hollinwilkins: Also, to be sure of the field that is being set, just do a quick:
timeCreated = LocalDateTime.now().toString

project/Dependencies.scala
```diff
@@ -8,7 +8,7 @@ object Dependencies {
 
   val sparkVersion = "2.1.0"
   val scalaTestVersion = "3.0.0"
-  val tensorflowVersion = "0.12.head"
+  val tensorflowVersion = "1.1.0"
```
hollinwilkins: This changeset should probably be in another PR

python/mleap/sklearn/pipeline.py
```diff
@@ -126,6 +126,7 @@ def get_bundle(transformer):
           "name": transformer.name,
           "format": "json",
           "version": __version__,
+          "timeCreated": datetime.datetime.now().isoformat(),
```
hollinwilkins: Again, just a note to double check that this returns iso8601 formatted date strings

bundle-ml/src/main/scala/ml/combust/bundle/dsl/Bundle.scala
jacobfoard: Yea they are both in the iso8601 format, and I have reflected the other two changes.

project/Dependencies.scala
jacobfoard: I'll open up another issue about this and put in the code I used for `getFields`

Issue #228, opened 2017-06-12
Getting error while running sbt compile


vidit-bhatia:
I am trying to build mleap using [documentation](http://mleap-docs.combust.ml/getting-started/building.html) . Unfortunately it is giving me errors in case of mlliblocal dependency.

'
Traceback (most recent call last):
  File "C:\Users\vibhatia\AppData\Local\Temp\protocbridge4017278385820396292.py"
, line 6, in <module>
    s.sendall(content)
TypeError: a bytes-like object is required, not 'str'
[info] Resolving jline#jline;2.12.1 ...
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  ::          UNRESOLVED DEPENDENCIES         ::
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]  :: org.apache.spark#spark-mllib-local_2.11;2.1.0: several problems occur
red while resolving dependency: org.apache.spark#spark-mllib-local_2.11;2.1.0 {c
ompile=[default(compile)]}:
[warn]  several problems occurred while resolving dependency: org.apache.spark#s
park-parent_2.11;2.1.0 {}:
[warn]  several problems occurred while resolving dependency: org.apache#apache;
14 {}:
[warn]  URI has an authority component
[warn]  URI has an authority component
[warn]
[warn]  URI has an authority component
[warn]
[warn]  URI has an authority component
[warn]  ::::::::::::::::::::::::::::::::::::::::::::::
[warn]
[warn]  Note: Unresolved dependencies path:
[warn]          org.apache.spark:spark-mllib-local_2.11:2.1.0 (E:\git\mleap\mlea
p-core\build.sbt#L4-5)
[warn]            +- ml.combust.mleap:mleap-core_2.11:0.8.0-SNAPSHOT
java.lang.RuntimeException: protoc returned exit code: 1
        at scala.sys.package$.error(package.scala:27)
        at sbtprotoc.ProtocPlugin$.sbtprotoc$ProtocPlugin$$compile(ProtocPlugin.
scala:114)
        at sbtprotoc.ProtocPlugin$$anonfun$sourceGeneratorTask$1$$anonfun$4.appl
y(ProtocPlugin.scala:141)
        at sbtprotoc.ProtocPlugin$$anonfun$sourceGeneratorTask$1$$anonfun$4.appl
'

Can someone please help me understand what am I doing wrong.

matw:
[...]

PR #239, opened 2017-07-14
update local maven repo with SBT internal representation

calinrc:
Fix for LeapFrameShow in case the frame contains null values.

Without the fix the below piece of code will fail with NPE: 

      val leapSchema = types.StructType(
        types.StructField("id", types.ByteType(true)),
        types.StructField("hour", types.ByteType(true)),
        types.StructField("mobile", types.FloatType(true)),
        types.StructField("userFeatures", types.TensorType(types.DoubleType(true), true)),
        types.StructField("clicked", types.DoubleType(true)),
        types.StructField("stringField", types.StringType(true)),
        types.StructField("stringField2", types.StringType(true))).get

      val validationData = LocalDataset(Array(ml.combust.mleap.runtime.Row(3.toByte, null, null, VectorConverters.sparkVectorToMleapTensor(Vectors.dense(2.0, 10.0, 0.5)), null, null, "testField2_4")))
      val frame = LeapFrame(leapSchema, validationData)
      frame.show()

Fix for Maven local repo client : 
If someone wants to build mleap project in windows using sbt command line will receive a failure with message "URI has an authority component". 
This can be fixed by adding double slash on file:// scheme
(e.g. "Local Maven Repository" at "file:////" + Path.userHome.absolutePath + "/.m2/repository", )
but this will fail on importing the project in some IDE (e.g. IntelliJ Idea ).
So the best way to fix both cases (sbt command line build and IDE import ) is to use sbt representation of local maven repo

### DIFF
--- a/mleap-runtime/src/main/scala/ml/combust/mleap/runtime/util/LeapFrameShow.scala
+++ b/mleap-runtime/src/main/scala/ml/combust/mleap/runtime/util/LeapFrameShow.scala
@@ -14,11 +14,11 @@ case class LeapFrameShow[LF <: LeapFrame[LF]](frame: LF, n: Int = 20, truncate: 
     val rows = schema.fields.map(_.name) +: dataset.take(n).toSeq.map {
       _.map {
         cell =>
-          val str = cell match {
+          val str = if (cell != null) cell match {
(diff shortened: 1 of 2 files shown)


calinrc:
This is fix for defect https://github.com/combust/mleap/issues/228

PR #249, opened 2017-08-29
Core Types for Python/Scikit/Pandas

seme0021:
@ancasarb Can you help describe the changes here?

### Transformers

Instead of `input_features` and `output_features` transformers' `mlinit` method now takes the FeatureExtractor object that was used to select the scalars or vectors that we need to transform.

The FeatureExtractor now has an `input_shapes` element that now get serialized to an mleap bundle.

### Feature Extractor

* Instead of input features, we now have to specify `input_scalars` or `input_vectors`
* It now generates an `input_shapes` element that gets used by the serialize to bundle method of the Transformers


hollinwilkins:
Closing this, as it was included in a separate PR.

PR #255, opened 2017-09-06
update spark and scikit-learn examples in readme(s)

ancasarb:


mleap-spark/src/main/resources/spark-demo.csv
```diff
@@ -0,0 +1,7 @@
+test_string,test_double
```
hollinwilkins: We probably shouldn't include this with the `mleap-spark` module. Can we create a folder next to `mleap-spark` called `examples` and add it here?

README.md
```diff
@@ -111,39 +111,46 @@ The first step is to create our pipeline in Spark. For our example we will manua
...
+import org.apache.spark.ml.feature.{Binarizer, StringIndexer}
+import org.apache.spark.sql._
+import org.apache.spark.sql.functions._
+import resource._
 
-// User out-of-the-box Spark transformers like you normally would
-val stringIndexer = new StringIndexerModel(uid = "si", labels = Array("hello", "MLeap")).
-  setInputCol("test_string").
-  setOutputCol("test_index")
+  val datasetName = "./mleap-docs/assets/spark-demo.csv"
```
hollinwilkins: This should probably point to the `examples` folder where we add the spark-demo.csv file.

ancasarb: Updated.

mleap-spark/src/main/resources/spark-demo.csv
ancasarb: Yep, good point, I intended to put it under test/resources. I've created the examples folder as you suggested instead.

PR #259, opened 2017-09-17
XGBoost Integration

hollinwilkins:
This pull request includes an implementation for XGBoost integration.

xgboost4j version: 0.7
https://github.com/komiya-atsushi/xgboost-predictor-java version: 0.2.1

We use the xgboost-predictor-java library so that the MLeap runtime does not need to rely on native xgboost libraries for evaluation.

Currently, we will only support the xgboost4j-spark provided classification and regression models. We may want to support the Spark models provided by xgboost-predictor-java as well eventually.


hollinwilkins:
Still need to add:
1. Parity specs for Spark integration
2. Runtime specs for models/transformers
3. Documentation and installation/usage guide
4. Maybe a Zeppelin notebook in mleap-docs that illustrates a complete pipeline

hollinwilkins:
Parity specs are done, still need to add in some runtime specs with a couple of different xgboost models

mleap-spark-testkit/src/main/scala/org/apache/spark/ml/parity/SparkParityBase.scala
```diff
@@ -79,14 +78,27 @@ abstract class SparkParityBase extends FunSpec with BeforeAndAfterAll {
...
+    val transformerInputTypes = transformer.exec.inputs.flatMap(_.dataTypes)
+
+    val modelOutputTypes = transformer.shape.outputs.
+      map(_._2.port).
+      map(n => model.outputSchema.getField(n).get.dataType).
+      toSeq
+    val transformerOutputTypes = transformer.exec.outputTypes
+
+    println(modelOutputTypes)
+    println(transformerOutputTypes)
```
ancasarb: printlns can be removed

project/MleapProject.scala
```diff
@@ -95,6 +101,21 @@ object MleapProject {
     dependencies = Seq(runtime)
   )
 
+  lazy val xgboost = Project(
+    id = "mleap-xgboost",
+    base = file("mleap-xgboost"),
+    dependencies = Seq(runtime)
+  )
+
+  lazy val xgboostSpark = Project(
```
ancasarb: Do you think we should include running these tests in travis.sh similarly to how we do the mleap-serving ones?

ancasarb:
I think that the build is failing because the order of the outputs in the shape in SupportVectorMachineOp is not correct, I think it needs to be rawPrediction->probability->prediction, as below.
```
  override def shape(node: SVMModel)(implicit context: BundleContext[SparkBundleContext]): NodeShape = {
    val rawPrediction = if(node.isDefined(node.rawPredictionCol)) Some(node.getRawPredictionCol) else None
    val probability = if(node.isDefined(node.probabilityCol)) Some(node.getProbabilityCol) else None

    NodeShape().withInput(node.getFeaturesCol, "features").
      withOutput(node.getRawPredictionCol, "raw_prediction").
      withOutput(node.getProbabilityCol, "probability").
      withOutput(node.getPredictionCol, "prediction")
  }
```

mleap-xgboost/src/main/scala/ml/combust/mleap/xgboost/bundle/ops/XGBoostRegressionOp.scala
```diff
@@ -0,0 +1,46 @@
...
+import resource._
+
+/**
+  * Created by hollinwilkins on 9/16/17.
+  */
+class XGBoostRegressionOp extends MleapOp[XGBoostRegression, XGBoostRegressionModel] {
+  override val Model: OpModel[MleapContext, XGBoostRegressionModel] = new OpModel[MleapContext, XGBoostRegressionModel] {
+    override val klazz: Class[XGBoostRegressionModel] = classOf[XGBoostRegressionModel]
+
+    override def opName: String = "xgboost.regression"
```
ancasarb: should we include these in Bundle.BuiltinOps like all others for consistency?

mleap-xgboost/src/main/scala/ml/combust/mleap/xgboost/bundle/ops/XGBoostClassificationOp.scala
```diff
@@ -0,0 +1,62 @@
...
+import resource.managed
+
+/**
+  * Created by hollinwilkins on 9/16/17.
+  */
+class XGBoostClassificationOp extends MleapOp[XGBoostClassification, XGBoostClassificationModel] {
+  override val Model: OpModel[MleapContext, XGBoostClassificationModel] = new OpModel[MleapContext, XGBoostClassificationModel] {
+    override val klazz: Class[XGBoostClassificationModel] = classOf[XGBoostClassificationModel]
+
+    override def opName: String = "xgboost.classifier"
```
ancasarb: as below, should probably include them in Bundle.BuiltinOps for consistency

hollinwilkins: I don't want to include these in builtin ops because they aren't part of mleap-spark or mleap-spark extension. xgboost is definitely a 3rd party integration outside the scope of standard transformers.

mleap-xgboost/src/main/scala/ml/combust/mleap/xgboost/bundle/ops/XGBoostRegressionOp.scala
hollinwilkins: See above

project/MleapProject.scala
hollinwilkins: It would be nice to run these in travis... Do you know of an easy way to include the xgboost shared library so we can run it in travis?

ancasarb: I'll take a look, but we can do this even after the PR gets merged.

mleap-xgboost-java/src/main/java/biz
```diff
@@ -0,0 +1 @@
+../../../../vendor/xgboost-predictor-java/xgboost-predictor/src/main/java/biz
```
hollinwilkins: This shows up kind of strange in github, but this is a symlink to the xgboost-predictor-java submodule src folder containing the predictor code.

mostafa-time:
Hey guys... not sure where to ask this question as stackoverflow did not have that many MLeap questions... Please advise: We use MLeap for deploying our ML engines to production (great job BTW! ).  I need to use an xgboost model in Spark ML pipeline, however it seems that xgboost4j-spark does not support "tree-method" -> "exact". So currently I use python xgboost for training. I was wondering wether I can convert the model obtained through python xgboost to XGBoostModel; i.e. do the training in python xgboost and convert it to XGBoostModel to use it with MLeap for scoring?

PR #260, opened 2017-09-18
Automatic casting of data types in leap frame and Spark DataFrame <--> LeapFrame conversion

hollinwilkins:
# Overview

There are two major changes in this pull request:

1. Automatic casting of data types when executing a LeapFrame
2. Conversion of Spark DataFrame <--> Spark LeapFrame

## Automatic Data Casting for LeapFrames

Because we now have very strict requirements on input/output types for all of our models, we have introduced automatic casting to cast data in a LeapFrame to the data expected by models/transformers. This takes into account nullable values as well as many common conversions between the basic data types, eg. int -> long, float -> int, etc.

Casting supports tensor -> tensor casting as well as list -> list, scalar -> tensor and tensor -> scalar

## Spark DataFrame <--> Spark LeapFrame

Instead of trying to support all of the different data types that could be found in a mixed Spark DataFrame that has been transformed with an MLeap transformer, we now generate an intermediate data structure called a SparkLeapFrame. An MLeap transformer can then operate on this data structure and all of the data types will be compatible with MLeap. After transformation, there is a reverse process in which the SparkLeapFrame converts back to a Spark DataFrame.


mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
```diff
@@ -0,0 +1,171 @@
...
+    (BasicType.Byte, BasicType.Boolean) -> { (v: Byte) => if (v != 0) true else false },
+    (BasicType.Byte, BasicType.Short) -> { (v: Byte) => v.toShort },
+    (BasicType.Byte, BasicType.Int) -> { (v: Byte) => v.toInt },
+    (BasicType.Byte, BasicType.Long) -> { (v: Byte) => v.toLong },
+    (BasicType.Byte, BasicType.Float) -> { (v: Byte) => v.toFloat },
+    (BasicType.Byte, BasicType.Double) -> { (v: Byte) => v.toDouble },
+    (BasicType.Byte, BasicType.String) -> { (v: Byte) => v.toString },
+
+    (BasicType.Short, BasicType.Boolean) -> { (v: Short) => if (v != 0) true else false },
+    (BasicType.Short, BasicType.Byte) -> { (v: Short) => v.toByte },
```
ancasarb: should below 

``` (BasicType.Short, BasicType.Int) -> { (v: Short) => v.toInt },
(BasicType.Short, BasicType.Long) -> { (v: Short) => v.toLong },
(BasicType.Short, BasicType.Float) -> { (v: Short) => v.toFloat },
(BasicType.Short, BasicType.Double) -> { (v: Short) => v.toDouble}
```

hollinwilkins: Fixed

PR #262, opened 2017-09-20
Handle null values like Spark does

hollinwilkins:
Spark handles null values through a mix of boxing/unboxing for primitive types and plain old null values for objects. We should handle nulls the same way.

This PR gets rid of the notion of an Option[T] as a null value in MLeap. Instead, we rely on boxing/unboxing and nulls.

This is desirable for several reasons:
1. It closer matches how Spark works
2. It will allow us to integrate more easily with libraries like Apache Arrow and Parquet
3. Java integration becomes much simpler, no need for handling Scala Options!

### DIFF
--- a/mleap-avro/src/main/scala/ml/combust/mleap/avro/DefaultFrameWriter.scala
+++ b/mleap-avro/src/main/scala/ml/combust/mleap/avro/DefaultFrameWriter.scala
@@ -31,7 +31,7 @@ class DefaultFrameWriter[LF <: LeapFrame[LF]](frame: LF) extends FrameWriter {
       for(row <- frame.dataset.toArray) {
         var i = 0
         for(writer <- writers) {
-          record.put(i, writer(row(i)))
+          record.put(i, writer(row.getRaw(i)))
           i = i + 1
         }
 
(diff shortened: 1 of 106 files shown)


ancasarb:
@hollinwilkins I was trying to add a Spark parity test for StringMap, and I modified StringMapModel schema to be 

``` 
  override def inputSchema: StructType = StructType("input" -> ScalarType.String.nonNullable).get

  override def outputSchema: StructType = StructType("output" -> ScalarType.Double.nonNullable).get

``` 
to match what it had before the changes in this PR, but it seems that a String is now always converted into a nullable ScalarType.String... is that intended, or am I missing something?

hollinwilkins:
@ancasarb Strings are different than the other primitive types because they are references. So it is easy to make an Int nullable or non-nullable. `Int` is non-nullable while `java.lang.Integer` is nullable because it is a boxed primitive. Strings are always nullable in the JVM, so it makes sense to have them be nullable by default. For the StringMap, you could explicitly set the types on the UDF to be non-nullable if that's desired.

hollinwilkins:
@ancasarb Can you open a PR with the StringMap changes you have linked? It would be great to get the parity spec in and we can work through the nullability stuff too.

ancasarb:
@hollinwilkins I've raised the PR for the StringMap changes https://github.com/combust/mleap/pull/277 which now include the parity test. I've left the input String as nullable, as Scala maps seem to allow it so I didn't quite see a reason not to. Let me know what you think.

RegexTokenizerModel and TokenizerModel now allow nullable inputs, but I think they would throw exceptions in the apply() methods, I'll raise another PR for that.

PR #268, opened 2017-09-22
Update mleap benchmark test models.

hollinwilkins:
This change updates the test models use to benchmark MLeap, this is needed because of the changes to the serialization format.


ancasarb:
@hollinwilkins I think the models in the S3 bucket used for mleap serving documentation under http://mleap-docs.combust.ml/mleap-serving/ also need updating.

Issue #272, opened 2017-09-26
MaxAbsScalerModel incorrect values when case SparseVector


smorring:
The MaxAbsScalerModel gives a SparseVector with different results than the equivalent model in Spark.

I think there are two problems:
line 40 of MaxAbsScalerModel.scala:
`vs(i) *= raw`

To make it emulate Spark the line would need to be:
`vs(i) = raw`

Also, line 38 of MaxAbsScalerModel.scala
`val raw = max(-1.0, min(1.0, values(i) / maxAbsUnzero(i)))`

To emulate Spark the line needs to be:
`val raw = max(-1.0, min(1.0, values(i) / maxAbsUnzero(indices(i))))`

I would've just made the code change for your review, but I've never used Git before.

hollinwilkins:
PR here: https://github.com/combust/mleap/pull/273

Thanks for finding this bug!

hollinwilkins:
@smorring Would you care to make a PR for review? Would be happy to help with any git knowledge you would need.

smorring:
I'd certainly love to contribute in some way.


I come from a SVN background, so I'm not familiar with Git.  I assume a PR is a Pull Request?


I've clicked "Create Pull Request" from GitHub and am presented with the following choices:

  *   release/v0.8<https://github.com/combust/mleap/compare/master...release/v0.8>14 minutes ago
  *    feature/ensemble<https://github.com/combust/mleap/compare/master...feature/ensemble>3 days ago
  *    feature/async<https://github.com/combust/mleap/compare/master...feature/async>3 days ago
  *    feature/cors<https://github.com/combust/mleap/compare/master...feature/cors>on Jun 30
  *    master@{1day}...master<https://github.com/combust/mleap/compare/master@%7B1day%7D...master>


I'm suppose I should click "release/v0.8?


FYI.  I'm using Intellij and can get to a Clone Repository dialog that asks for a "Git Repository URL".


I'd love to contribute if you could help me get started the first time.


________________________________

hollinwilkins:
1. First thing to do is to fork mleap. Click on the button in the upper right on this screen. This makes a copy of mleap for your user.

2. Next thing is to clone the fork.
`git clone --recursive git@github.smorring/mleap.git`
This makes a local copy on your computer.

3. Make a branch to work on `git branch feature/issue272 && git checkout feature/issue272`

4. Make your changes

5. Commit your changes `git add . && git commit -m "Fix issue with MaxAbsScaler"`

[...]

PR #273, opened 2017-09-26
Fix MaxAbsScaler.

hollinwilkins:
This fixes the issue with MaxAbsScaler not producing the same values as Spark.

https://github.com/combust/mleap/issues/272


smorring:
FYI.  I edited the initial issue to include a needed change to line 38 which doesn't appear to be in the commit.
Line 38 needs to be: 
`val raw = max(-1.0, min(1.0, values(i) / maxAbsUnzero(indices(i))))`

Without the "indices" the wrong scale is being applied to non-consecutive indices.

hollinwilkins:
@smorring Would you like to generate a PR for this issue for review? It would be a good introduction to git :)

PR #278, opened 2017-09-28
Mark input to RegexTokenizer and Tokenizer models as nonnullable to prevent NPEs

ancasarb:


mleap-runtime/src/main/scala/ml/combust/mleap/runtime/transformer/feature/RegexTokenizer.scala
```diff
@@ -8,5 +8,6 @@ import ml.combust.mleap.runtime.transformer.{SimpleTransformer, Transformer}
 case class RegexTokenizer(override val uid: String = Transformer.uniqueName("regex_tokenizer"),
                           override val shape: NodeShape,
                           override val model: RegexTokenizerModel) extends SimpleTransformer {
-  override val exec: UserDefinedFunction = (value: String) => model(value)
+  override val exec: UserDefinedFunction = UserDefinedFunction((value: String) => model(value),
```
hollinwilkins: This isn't strictly necessary because typedExec in SimpleTransformer will use the types from the model for the UDF. Still nice to have it be explicit though.

ancasarb: The parity tests would fail otherwise, but I think I understood why. https://github.com/combust/mleap/pull/280 The typedExec would only use the output types from the model for the UDF but not the input, I don't think.

PR #282, opened 2017-10-02
Allow serialization format to be set in SimpleSparkSerializer

clocklear:
This simple PR allows the bundle serialization format to be set in `SimpleSparkSerializer`.  The default is Json, much like everywhere else, so the change is backwards compatible.


clocklear:
I have not been able to test it, as we don't use the PySpark integration.  A cursory glance of the python code doesn't reveal any uses of `SimpleSparkSerializer`, am I missing something?

hollinwilkins:
Here are the references from PySpark.

https://github.com/combust/mleap/blob/master/python/mleap/pyspark/spark_support.py#L36-L45

hollinwilkins:
My only concern is that the method call won't work because there is an extra parameter now, even though it does have a default.

clocklear:
I'll see if I can figure out how to (at very least) set a new param on the python side that defaults to Json.

hollinwilkins:
@clocklear A suggestion for a safer change (no Python needed), add an entirely new method that takes a SerializationFormat as a parameter. Have the existing method call that one with a default of SerializationFormat.Json.

PR #283, opened 2017-10-02
Treat UNDEFINED as empty value for BasicType

clocklear:
Note that this is related to [this bundle-protobuf PR](https://github.com/combust/bundle-protobuf/pull/1) (and requires that PR to be merged first).

Per protobuf best practices (here's an example of the problem: http://androiddevblog.com/protocol-buffers-pitfall-adding-enum-values/), you should never use '0' as a real value in your enums.  This would prevent you from being able to determine if the value is 'empty' or 'default', as the language generated code always uses the value at 0 as the default for the field.

This is causing problems in mleap for BasicType serialization in transformers.  If the type being serialized is `BasicType.BOOLEAN`, the code has no way to determine if the value has been set or is the default, which is causing mleap to skip serialization of the field.

Please note that this is entirely NOT backwards compatible (as in Mleap models previously serialized as protobuf will no longer deserialize properly with regards to `BasicType`).  It is, however, the correct solution in this scenario (in my opinion).


hollinwilkins:
@clocklear I just merged the bundle-protobuf changes in, you will have to commit the git submodule update for this PR to pass travis CI.

PR #284, opened 2017-10-04
Initial pass at TfidfVectorizer

mslipper:


python/mleap/sklearn/feature_extraction/text.py
```diff
@@ -15,31 +15,49 @@
...
 setattr(CountVectorizer, 'op', 'tokenizer')
-setattr(CountVectorizer, 'mlinit', mleap_init)
-setattr(CountVectorizer, 'serialize_to_bundle', serialize_to_bundle)
+setattr(CountVectorizer, 'mlinit', count_vectorizer_mleap_init)
+setattr(CountVectorizer, 'serialize_to_bundle', count_vectorizer_serialize_to_bundle)
 setattr(CountVectorizer, 'serializable', True)
 
 
-class SimpleSerializer(MLeapSerializer):
+def tfidf_vectorizer_serialize_to_bundle(self, path, model_name):
```
ancasarb: Perhaps you can take a look at https://github.com/combust/mleap/blob/master/python/mleap/sklearn/preprocessing/data.py#L74 and https://github.com/combust/mleap/blob/master/python/mleap/sklearn/preprocessing/data.py#L155 and see what you think about how we've made mleap_init and serialize_to_bundle to support multiple serializers and maybe follow the same pattern?

python/mleap/sklearn/feature_extraction/text.py
```diff
@@ -65,4 +83,74 @@ def serialize_to_bundle(self, transformer, path, model_name):
...
+            ('binary', transformer.binary),
+            # no concept of 'min_tf' in scikit's TfidfVectorizer
+            ('min_tf', None)
+        ]
+        tf_inputs = [{
+            "name": transformer.input_features,
+            "port": "input"
+        }]
+        tf_outputs = [{
+            "name": "{}_count_vectorizer".format(transformer.name),
```
ancasarb: I think we can just put the output name of the count vectorizer to be something like 'token_counts' or something similar... it'll make it a bit easier to read if it doesn't have the transformer name in it

python/mleap/sklearn/feature_extraction/text.py
```diff
@@ -65,4 +83,74 @@ def serialize_to_bundle(self, transformer, path, model_name):
...
+        vocabulary = [None] * num_features
+
+        for term, termidx in transformer.vocabulary_.iteritems():
+            vocabulary[termidx] = str(term)
+
+        tf_attributes = [
+            ('vocabulary', vocabulary),
+            ('binary', transformer.binary),
+            # no concept of 'min_tf' in scikit's TfidfVectorizer
+            ('min_tf', None)
```
ancasarb: Have you tried to use the serialized models to transform some leap frames? 

I think it'll likely fail if it doesn't have min_tf currently, I'll have a look what changes are needed to get the transform to work without min_tf.

ancasarb:
I'm not sure what changed for file run-tests.sh, is it just Github showing it funny?

python/mleap/sklearn/feature_extraction/text.py
mslipper: Will do!

python/mleap/sklearn/feature_extraction/text.py
mslipper: I'll take a look at transforming leap frames. I'll also do a deep dive into the scikit code and see if there's a way I can infer `min_tf`.

python/mleap/sklearn/feature_extraction/text.py
mslipper: Sounds good!

mslipper:
I made `run-tests.sh` executable.

python/mleap/sklearn/feature_extraction/text.py
```diff
@@ -15,33 +15,64 @@
...
+    def __init__(self):
+        self.COUNT_VECTORIZER = 'tokenizer'
+        self.TF_IDF_VECTORIZER = 'pipeline'
+
+
+ops = ops()
+
+
 def serialize_to_bundle(self, path, model_name):
-    serializer = SimpleSerializer()
```
ancasarb: I believe this should be SimpleSerializer so that CountVectorizer would work on its own and not just part of the tf-idf.

python/mleap/sklearn/feature_extraction/text.py
```diff
@@ -65,4 +105,74 @@ def serialize_to_bundle(self, transformer, path, model_name):
...
+            ('binary', transformer.binary),
+            # no concept of 'min_tf' in scikit's TfidfVectorizer
+            ('min_tf', 0)
+        ]
+        tf_inputs = [{
+            "name": transformer.input_features,
+            "port": "input"
+        }]
+        tf_outputs = [{
+            "name": "token_counts".format(transformer.name),
```
ancasarb: .format() is no longer needed here.

python/mleap/sklearn/feature_extraction/text.py
```diff
@@ -65,4 +105,74 @@ def serialize_to_bundle(self, transformer, path, model_name):
...
+            "port": "output"
+        }]
+        tf_step = TfidfStep(transformer, 'count_vectorizer', tf_attributes, tf_inputs, tf_outputs)
+
+        idf_attributes = [
+            ('idf', transformer.idf_)
```
ancasarb: I think it needs to be transformer.idf_.tolist() for the serialized model to work correctly with mleap serving.

python/mleap/sklearn/feature_extraction/tests.py
```diff
@@ -0,0 +1,84 @@
...
+
+        idf_model_file = '{}/{}/root/{}.node/model.json'.format(self.tmp_dir, self.tfidf.name, idf_node_name)
+        with open(idf_model_file) as idf_model_json:
+            idf_model = json.load(idf_model_json)
+
+        idf_shape = {
+            'dimensions': [
+                {
+                    'name': '',
+                    'size': [1]
```
ancasarb: should be 'size': 1 after the idf_attributes changes.

python/mleap/sklearn/feature_extraction/text.py
mslipper: Nice catch. Will fix, and write a test for CountVectorizer.

mslipper:
PR updated. Quick question, though - is `CountVectorizer` supposed to be serialized as a tokenizer? This is what the original code in `text.py` does, want to make sure that it's not a bug. If it is a bug, I'd be happy to fix it once this PR gets merged.

mslipper:
Any time :)

PR #287, opened 2017-10-09
Feature/pypi version

seme0021:
Updating version to `0.8.1`


seme0021:
Not sure what all of these other commits are - @hollinwilkins thoughts?

hollinwilkins:
@seme0021 Not sure why so many commits, will squash them into one when I merge it in though.

ancasarb:
Thank you!

PR #290, opened 2017-10-11
Pipeline schema updates

ancasarb:


hollinwilkins:
This looks good. Calculating the input/output schema has been complicated a bit by adding in drop and select to the frame builder. We may want to go back to using a frame builder that calculates input/output schema and takes into account drops/selects.

ancasarb:
I was thinking that because of the filtering on the Pipeline inputs but not outputs, there wasn't a way to retrieve the actual outputs for a pipeline, which I think would be quite useful. 

But I found in the meantime that my original solution wasn't working very well with nested pipelines, when I tried it with the airbnb example, it was giving funky results about what was an actualInput or actualOutput and not. 

I think that doing any filtering on the Pipeline itself may be a bit too early and so, I've made the changes to reflect this. If we need to do any filtering to get the actualInputs/actualOutputs, I was thinking a better place to do it would be in MleapService in the mleap-serving module and any mleap client can implement the filtering as best suited.

mleap-runtime/src/main/scala/ml/combust/mleap/runtime/transformer/Pipeline.scala
```diff
@@ -47,7 +47,7 @@ case class Pipeline(override val uid: String = Transformer.uniqueName("pipeline"
           oacc ++ tform.outputSchema.fields.map(f => f.name -> f.dataType))
     }
 
-    val actualInputs = (inputs -- outputs.keys).map {
+    val actualInputs = inputs.map {
```
hollinwilkins: What is the problem with getting rid of outputs? I would think that is the right thing to do?

ancasarb: I was thinking of a pipeline that looks like this (may not be too realistic in Spark, but the scikit pipelines look a lot like this. And what I wanted to get out of the outputSchema is the actual output of the root pipeline, i.e. just the 'scaled_features' field. This is not actually possible at the moment with the current implementation of inputSchema/outputSchema of the Pipeline.

``` val pipeline: Transformer = new Pipeline().setStages(Array(
    new Pipeline().setStages(Array(
    new VectorAssembler().
    setInputCols(Array("dti", "loan_amount")).
    setOutputCol("features"))),
      new Pipeline().setStages(Array(
    new StandardScaler().
      setInputCol("features").
      setOutputCol("scaled_features"))))) ```

hollinwilkins: Here is the way I think of input schema and output schema:

Input schema: All of the fields that must be in the leap frame for the transformer to execute
Output schema: All of the fields that are added to the leap frame after the transformer executes

Maybe we should actually have 3 schemas though:
1. Input schema: All of the fields that must be in the leap frame for the transformer to execute
2. Intermediate schema: All of the fields that are both inputs/outputs to some stage of the transformer
3. Output schema: All of the fields that are added to the leap frame that are not intermediate fields

ancasarb: ok, that makes sense to me. is it ok if I make those changes? we only really need three schemas for the pipeline right?

hollinwilkins: Go ahead and make the changes, yeah I think we only need 3 for pipeline. We may also want to add a fourth:

4. Strict output schema: outputSchema -- intermediateSchema

ancasarb: Sounds good.

ancasarb:
Still need to check with the airbnb schema that it works fine.

ancasarb:
@hollinwilkins I've checked now that the schemas look file with the airbnb model and a couple more scikit models.

PR #300, opened 2017-11-13
issue 231 - support sparse data in standard scaler

ancasarb:


ancasarb:
@hollinwilkins Good point, I looked at Spark code and it looks like they are treated as zeros so the mean gets subtracted from 0 for every unset value. I've updated the standard scaler code with the one to match Spark (which supports sparse vectors now), let me know what you think.
https://github.com/apache/spark/blob/39e2bad6a866d27c3ca594d15e574a1da3ee84cc/mllib/src/main/scala/org/apache/spark/mllib/feature/StandardScaler.scala

PR #333, opened 2018-02-13
Update BundleFile to handle '\' in Files and URIs

aufrank:
When mleap is built in Windows, operations related to getting a string from a File, Path, or URI will include `\` characters.  This includes calls to:
 * `file.toString()`
 * `file.getPath()`
 * `file.toPath().toString()`
 * `file.toPath().toUri().toString()`
 * `file.toURI().toString()`

To be defensive, we create new URIs from all inputs, with backslashes replaced by forward slashes.

### DIFF
--- a/bundle-ml/src/main/scala/ml/combust/bundle/BundleFile.scala
+++ b/bundle-ml/src/main/scala/ml/combust/bundle/BundleFile.scala
@@ -23,6 +23,7 @@ object BundleFile {
     apply(new URI(unbackslash(uri)))
   }
 
+
   implicit def apply(file: File): BundleFile = {
     val uri: String = if (file.getPath.endsWith(".zip")) {
       s"jar:${file.toURI.toString}"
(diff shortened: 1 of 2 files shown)


bundle-ml/src/main/scala/ml/combust/bundle/BundleFile.scala
```diff
@@ -20,32 +20,43 @@ import scala.util.Try
...
         (zfs, zfs.getPath("/"))
     }
 
     apply(fs, path)
   }
+
+  private def unbackslash(uri: String): String = {
+    // same logic as https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/FilenameUtils.java#L570
+    val BACKSLASH = '\\'
+    val NOT_FOUND = -1
```
hollinwilkins: We shouldn't check for null URI, it should fail because this is definitely not expected input. Also, don't bother first checking for if backslash is in the string or not, just perform the replace.

PR #337, opened 2018-02-28
Fix load from resource

massivityreport:
This fix handles bundle loaded from a resourse inside a jar.

e.g: 
  private val sourceURI = Thread.currentThread().getContextClassLoader.getResource("models/similar-proposals-rf").toURI
  private val bundle = BundleFile(sourceURI).load().get
=>
jar:file:/tmp/sbt_1ccf9912/job-1/target/d8dd53b3/semantic_2.11-0.1.6.jar!/models/similar-proposals-rf

### DIFF
--- a/bundle-ml/src/main/scala/ml/combust/bundle/BundleFile.scala
+++ b/bundle-ml/src/main/scala/ml/combust/bundle/BundleFile.scala
@@ -42,8 +42,17 @@ object BundleFile {
       case "file" =>
         (FileSystems.getDefault, FileSystems.getDefault.getPath(uriSafe.getPath))
       case "jar" =>
-        val zfs = FileSystems.newFileSystem(uriSafe, env)
-        (zfs, zfs.getPath("/"))
+        // handle resource in JAR path
+        val (filesystemUri: URI, path: String) = if (uriSafe.toString.contains("!")) {
+          val uriParts: Array[String] = uriSafe.toString.split("!")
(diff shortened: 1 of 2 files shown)


bundle-ml/src/main/scala/ml/combust/bundle/BundleFile.scala
```diff
@@ -42,8 +42,17 @@ object BundleFile {
       case "file" =>
         (FileSystems.getDefault, FileSystems.getDefault.getPath(uriSafe.getPath))
       case "jar" =>
-        val zfs = FileSystems.newFileSystem(uriSafe, env)
-        (zfs, zfs.getPath("/"))
+        // handle resource in JAR path
```
hollinwilkins: Can you add a test for this to make sure we aren't breaking anything?

hollinwilkins:
I like this idea a lot to include bundles in the jar files. Makes it really easy to distribute the models as JVM libraries.

bundle-ml/src/main/scala/ml/combust/bundle/BundleFile.scala
massivityreport: Sure. Added.

massivityreport:
It's very practical. Right now it only works with a directory though. It would be nice to make it work with a zipped bundle in the jar.

massivityreport:
Test added as requested.

PR #338, opened 2018-02-28
Fix scikit-learn Random Forest serialization

massivityreport:
The number of classes was read from the wrong attribute of the RandomForestClassifier and of the DecisionTreeClassifier. 

As a result Mleap was unable to load and run the bundle.

### DIFF
--- a/python/mleap/sklearn/ensemble/forest.py
+++ b/python/mleap/sklearn/ensemble/forest.py
@@ -89,7 +89,7 @@ class SimpleSerializer(MLeapSerializer):
         attributes.append(('num_features', transformer.n_features_))
         attributes.append(('tree_weights', tree_weights))
         attributes.append(('trees', ["tree{}".format(x) for x in range(0, len(transformer.estimators_))]))
-        if transformer.n_classes_ > 1:
+        if isinstance(transformer, RandomForestClassifier):
             attributes.append(('num_classes', transformer.n_classes_)) # TODO: get number of classes from the transformer
 
         self.serialize(transformer, path, model, attributes, inputs, outputs)
--- a/python/mleap/sklearn/ensemble/forest.py
+++ b/python/mleap/sklearn/ensemble/forest.py
@@ -89,8 +89,8 @@ class SimpleSerializer(MLeapSerializer):
         attributes.append(('num_features', transformer.n_features_))
         attributes.append(('tree_weights', tree_weights))
         attributes.append(('trees', ["tree{}".format(x) for x in range(0, len(transformer.estimators_))]))
-        if transformer.n_outputs_ > 1:
-            attributes.append(('num_classes', transformer.n_outputs_)) # TODO: get number of classes from the transformer
+        if transformer.n_classes_ > 1:
+            attributes.append(('num_classes', transformer.n_classes_)) # TODO: get number of classes from the transformer
 
         self.serialize(transformer, path, model, attributes, inputs, outputs)
 
--- a/python/mleap/sklearn/tree/tree.py
+++ b/python/mleap/sklearn/tree/tree.py
@@ -114,8 +114,8 @@ class SimpleSerializer(MLeapSerializer):
         # Define attributes
         attributes = list()
         attributes.append(('num_features', transformer.n_features_))
-        if transformer.classes_ is not None:
-            attributes.append(('num_classes', transformer.n_classes_))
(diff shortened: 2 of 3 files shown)


python/mleap/version.py
```diff
@@ -16,4 +16,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.8.1"
+__version__ = "0.8.1-fix"
```
hollinwilkins: Can you actually bump version to 0.9.4?

massivityreport: sure :) done

python/mleap/sklearn/tree/tree.py
```diff
@@ -114,8 +114,8 @@ def serialize_to_bundle(self, transformer, path, model_name, serialize_node=True
         # Define attributes
         attributes = list()
         attributes.append(('num_features', transformer.n_features_))
-        if transformer.classes_ is not None:
-            attributes.append(('num_classes', transformer.n_classes_))
+        if transformer.n_classes_ is not None:
```
ancasarb: we were using classes_ to determine when to append the num_classes to the attributes because that is not None when using a DecisionTreeClassifier but None when we have a DecisionTreeRegressor. perhaps it would be better to just check which type of transformer we're dealing with instead. what do you think?

python/mleap/sklearn/ensemble/forest.py
```diff
@@ -89,8 +89,8 @@ def serialize_to_bundle(self, transformer, path, model):
         attributes.append(('num_features', transformer.n_features_))
         attributes.append(('tree_weights', tree_weights))
         attributes.append(('trees', ["tree{}".format(x) for x in range(0, len(transformer.estimators_))]))
-        if transformer.n_outputs_ > 1:
-            attributes.append(('num_classes', transformer.n_outputs_)) # TODO: get number of classes from the transformer
+        if transformer.n_classes_ > 1:
```
ancasarb: what was the failure you were seeing?

i'm asking because we have this example https://github.com/combust/mleap-demo/blob/master/notebooks/airbnb-price-regression-scikit.ipynb that includes a random forest as well, and i could score the serialized model with mleap-scoring.

ancasarb: I see now that that the demo uses a RandomForestRegressor instead of a RandomForestClassifier.

python/mleap/version.py
```diff
@@ -16,4 +16,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.8.1"
+__version__ = "0.9.4"
```
ancasarb: @hollinwilkins maybe we could release an updated pip package once this gets merged.

python/mleap/sklearn/tree/tree.py
massivityreport: It would be way better to check the type of the classifier. classes_ handles a special case when the dimension of the output is greater than one (multi-output problem). I have never used it.

python/mleap/sklearn/ensemble/forest.py
massivityreport: I was indeed for a RandomForestClassifier. The  regressor has no classes and thus no issues.

ancasarb: okay, that makes sense. if it's ok with you, could you please update here the if statement as well to check the type of the transformer rather than the n_classes?

python/mleap/sklearn/tree/tree.py
ancasarb: sounds good, would it be ok to update the if statement then to check the type of the transformer?

python/mleap/sklearn/ensemble/forest.py
massivityreport: sure. I'll do that.

massivityreport: Sorry, I was a bit busy, but I have updated the check and pushed it.

massivityreport:
Statement updated as requested.

ancasarb:
looks good to me, @massivityreport thanks for making the changes!

hollinwilkins:
@ancasarb Ready to merge this in, but need you to approve the changes requested.

ancasarb:
@hollinwilkins approved :)

Issue #307, opened 2017-11-22
Mleap runtime Logistic regression thresholds


wanchaol:
Hi, I tested locally by training the logistic regression model, serliaizing to disk, then deserialize and make prediction, the logistic regression in the runtime is having the "thresholds" variable empty. It probably because you are not setting the thresholds inside the ProbablisticClassificationModel using the impl. 

```scala
case class LogisticRegressionModel(impl: AbstractLogisticRegressionModel) extends ProbabilisticClassificationModel {
  override val numClasses: Int = impl.numClasses
  override val numFeatures: Int = impl.numFeatures
  val isMultinomial: Boolean = impl.numClasses > 2

  def multinomialModel: ProbabilisticLogisticsRegressionModel = impl.asInstanceOf[ProbabilisticLogisticsRegressionModel]
  def binaryModel: BinaryLogisticRegressionModel = impl.asInstanceOf[BinaryLogisticRegressionModel]

  override def predict(features: Vector): Double = impl.predict(features)

  override def predictRaw(features: Vector): Vector = impl.predictRaw(features)

  override def rawToProbabilityInPlace(raw: Vector): Vector = impl.rawToProbabilityInPlace(raw)
}
```

wanchaol:
I think it also perhaps the problem is related to calling predictRaw instead of predict

ancasarb:
Hi @wanchaol, 

I am trying to reproduce the error and I was wondering what did you use to serialize the model, Spark, MLeap? Are the "thresholds" serialized to the bundle but not set when the model is loaded at scoring time?

https://github.com/combust/mleap/blob/master/mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/classification/LogisticRegressionOp.scala#L48

I can see the "thresholds" being set in the LogisticRegressionOp but perhaps there's an issue and it doesn't get set properly.

wanchaol:
Hi, @ancasarb I trained using spark, and serialize it to bundle, the bundle contains the thresholds, but when I try deserializing and do prediction, the threshold is not actually applied to the prediction

hollinwilkins:
@wanchaol This is probably an issue related to a discussion we have been having in gitter. There are probably a few Spark deserializers that are not getting all of the information out of the bundle back into Spark.

ancasarb:
@hollinwilkins @wanchaol I've been working on a fix for this. 

[...]

PR #342, opened 2018-03-18
fix deserializing multinomial logistic regression in spark

ancasarb:
Fix for issue #307.

### DIFF
--- a/mleap-core/src/main/scala/ml/combust/mleap/core/classification/LogisticRegressionModel.scala
+++ b/mleap-core/src/main/scala/ml/combust/mleap/core/classification/LogisticRegressionModel.scala
@@ -46,22 +46,6 @@ case class ProbabilisticLogisticsRegressionModel(coefficientMatrix: Matrix,
   override val numClasses: Int = interceptVector.size
   override val numFeatures: Int = coefficientMatrix.numCols
 
-  lazy val binaryModel: BinaryLogisticRegressionModel = {
-    val coefficients = {
-      require(coefficientMatrix.isTransposed,
-        "LogisticRegressionModel coefficients should be row major.")
-      coefficientMatrix match {
-        case dm: DenseMatrix => Vectors.dense(dm.values)
-        case sm: SparseMatrix => Vectors.sparse(coefficientMatrix.numCols, sm.rowIndices, sm.values)
-      }
-    }
-    val intercept = interceptVector(0)
-
-    BinaryLogisticRegressionModel(coefficients = coefficients,
-      intercept = intercept,
-      threshold = thresholds.get(0))
-  }
-
   private def margins(features: Vector): Vector = {
     val m = interceptVector.toDense.copy
     BLAS.gemv(1.0, coefficientMatrix, features, 1.0, m)
--- a/mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/classification/LogisticRegressionOpV21.scala
+++ b/mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/classification/LogisticRegressionOpV21.scala
@@ -43,7 +43,7 @@ class LogisticRegressionOpV21 extends SimpleSparkOp[LogisticRegressionModel] {
         val coefficientMatrix = Matrices.dense(cmTensor.dimensions.head, cmTensor.dimensions(1), cmTensor.toArray)
         val lr = new LogisticRegressionModel(uid = "",
           coefficientMatrix = coefficientMatrix,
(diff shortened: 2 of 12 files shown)


mleap-core/src/main/scala/ml/combust/mleap/core/classification/LogisticRegressionModel.scala
```diff
@@ -46,22 +46,6 @@ case class ProbabilisticLogisticsRegressionModel(coefficientMatrix: Matrix,
   override val numClasses: Int = interceptVector.size
   override val numFeatures: Int = coefficientMatrix.numCols
 
-  lazy val binaryModel: BinaryLogisticRegressionModel = {
```
ancasarb: @hollinwilkins this doesn't seem to be used anywhere, I decided to remove it.

mleap-spark/src/test/scala/org/apache/spark/ml/parity/classification/MultinomialLogisticRegressionParitySpec.scala
```diff
@@ -0,0 +1,34 @@
...
+    .add("height", IntegerType, nullable = false)
+    .add("weight", IntegerType, nullable = false)
+
+  override val dataset: DataFrame = spark.sqlContext.createDataFrame(rows, schema)
+
+  override val sparkTransformer: Transformer = new Pipeline().setStages(Array(
+    new VectorAssembler().
+      setInputCols(Array("age", "height", "weight")).
+      setOutputCol("features"),
+    new LogisticRegressionModel(uid = "logr",
```
ancasarb: @hollinwilkins I ended up using LogisticRegressionModel directly for the parity test, because I saw the same in the existing LogisticRegressionParitySpec.scala parity test, which tests the binary case. 

I initially tried to use LogisticRegression instead and as far as I saw, the coefficientMatrix/interceptVector for the LogisticRegressionModel created were the same, but the equalityTest() was failing, because the results were ever so slightly different. Do you happen to know why that would be?

hollinwilkins: @ancasarb I wonder if the values are being stored as JSON and losing some precision? Or losing precision in protobuf?

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/classification/LogisticRegressionOpV21.scala
```diff
@@ -43,7 +43,7 @@ class LogisticRegressionOpV21 extends SimpleSparkOp[LogisticRegressionModel] {
         val coefficientMatrix = Matrices.dense(cmTensor.dimensions.head, cmTensor.dimensions(1), cmTensor.toArray)
         val lr = new LogisticRegressionModel(uid = "",
           coefficientMatrix = coefficientMatrix,
-          interceptVector = Vectors.dense(model.value("intercept").getTensor[Double].toArray),
+          interceptVector = Vectors.dense(model.value("intercept_vector").getTensor[Double].toArray),
```
hollinwilkins: Can we leave this as "intercept" so we don't break our serialization format?

mleap-core/src/main/scala/ml/combust/mleap/core/classification/LogisticRegressionModel.scala
hollinwilkins: Probably unneeded. I may have just taken it from Spark.

hollinwilkins:
@ancasarb Just a couple of things to sort then let's merge this

mleap-spark/src/test/scala/org/apache/spark/ml/parity/classification/MultinomialLogisticRegressionParitySpec.scala
ancasarb: @hollinwilkins the values stored for coefficients/intercepts seemed the same, as far as i could tell. but yea, suspect something similar is happening.

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/classification/LogisticRegressionOpV21.scala
ancasarb: @hollinwilkins in case of multinomial log regr, it is being stored as "intercept_vector", here is just the fix to load it back correctly.

hollinwilkins:
@ancasarb I'm good to merge this, just want to check in with you first

ancasarb:
@hollinwilkins i'm ok to merge this too.

PR #344, opened 2018-03-21
.im_class has been removed in Python 3. Use __class__.__name__ instead

trams:
I tried your demo (https://github.com/combust/mleap-demo/blob/master/notebooks/airbnb-price-regression-scikit.ipynb) and I had troubles running it in Python 3 even though mleap library supports Python 3 according to setup.py

It seems there was only one place where you use something Python 2 specific. I suggest this modification to make it Python 3 compatible (it still works on Python 2)

I ran unit tests and everything seems ok

### DIFF
--- a/python/mleap/sklearn/preprocessing/data.py
+++ b/python/mleap/sklearn/preprocessing/data.py
@@ -87,7 +87,7 @@ def mleap_init(self, prior_tf, output_features=None):
         output_size = 1
         output_feature_name = prior_tf.output_features
 
-    class_name = "{}".format(self.__init__.im_class).split('.')[-1].replace('>','').replace("'",'')
+    class_name = self.__class__.__name__
 
     if output_features is not None:
         self.output_features = output_features


python/mleap/bundle/serialize.py
```diff
@@ -16,6 +16,13 @@ def __init__(self, values):
         self.values = values
 
 
+class EncoderNumpySupport(json.JSONEncoder):
```
ancasarb: @trams I see there is another PR that contains these changes, could you please remove them from this one?

python/mleap/sklearn/preprocessing/data.py
```diff
@@ -87,7 +87,7 @@ def mleap_init(self, prior_tf, output_features=None):
         output_size = 1
         output_feature_name = prior_tf.output_features
 
-    class_name = "{}".format(self.__init__.im_class).split('.')[-1].replace('>','').replace("'",'')
+    class_name = "{}".format(self.__class__.__name__).split('.')[-1].replace('>','').replace("'",'')
```
ancasarb: @trams with these changes, I believe the following will suffice. the result from ```im_class``` was in the format ``` <class 'mleap.sklearn.preprocessing.data.FeatureExtractor'> ``` and that's why there were the extra processing steps afterwards.

``` 
class_name = self.__class__.__name__
```

trams: Yes. You are right. I'll submit a new patch

python/mleap/bundle/serialize.py
trams: Yes. I divided this into two pull requests

PR #345, opened 2018-03-21
Add support for numpy numbers to be json serializable

trams:
The problem was discovered by running
https://github.com/combust/mleap-demo/blob/master/notebooks/airbnb-price-regression-scikit.ipynb

During a bundle serialization an object was constructed
to describe a model. This object for some reason contained numpy numbers
instead of python native integers.

Therefore it was not json serializable cause the type was not recognized

This patch adds support for numpy numbers for json serializer

### DIFF
--- a/python/mleap/bundle/serialize.py
+++ b/python/mleap/bundle/serialize.py
@@ -16,6 +16,13 @@ class Vector(object):
         self.values = values
 
 
+class EncoderNumpySupport(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, np.number):
+            return np.asscalar(obj)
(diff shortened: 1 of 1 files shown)


ancasarb:
@trams could you please give a bit more details on which transformer or part of the pipeline was having this issue for the airbnb example? thank you!

hollinwilkins:
@ancasarb @trams Is there an update on this? Should we merge this in?

ancasarb:
@hollinwilkins I was just trying to understand what the issue that's being fixed is. as far as i am aware, we can serialize and score the airbnb without any issues.

ancasarb:
Closing this, due to inactivity, please re-open if it's still an issue.

Issue #325, opened 2018-01-16
Allow user to set the listening port for mleap serving as command line argument


davidecoccia:
I would like to have the possibility to set the listening port from the starting command (as a parameter) so I will be able to launch multiple instances of mleap serving on the same machine without using docker

polya20:
I just had to change mleap serving code  to run mleap serving on aws sagemaker

hollinwilkins:
@davidecoccia We will be overhauling MLeap serving as part of 1.0. Putting together a roadmap soon and posting it in the README. Will close this issue and point you there when it is ready.

cappaberra:
@davidecoccia, I opened pull request https://github.com/combust/mleap/pull/366 that consequently helps with this request. Hopefully we can get it reviewed / discussed / merged soon. :)

hollinwilkins:
Merged #366

PR #366, opened 2018-05-19
MLeap Serving Parameters

cappaberra:
Motivation for this PR:  I needed a convenient way to manage what interface and port the MLeap serving binary listens on in order to facilitate an automated approach to having multiple MLeap serving instances running on the same server (e.g. by using different ports).  The primary objective is to host multiple instances each with their own unique model.

A few disclaimers:
- I am NOT a Scala developer! This is my first foray in it... :)
- The port/bind-port and hostname/bind-hostname variables seemed redundant? I removed the one that wasn't used by the MleapServer.scala code.  I didn't see any evidence that the "redundant" variables were used anywhere else, so please correct me if I'm wrong there.
- Just realized.... prior to doing Http().bindAndHandle(), the variables (maybe?) should be validated for at least format, since they could have been passed in via the shell environment.

Thanks!

### DIFF
--- a/mleap-serving/src/main/resources/application.conf
+++ b/mleap-serving/src/main/resources/application.conf
@@ -1,6 +1,8 @@
 ml.combust.mleap.serving {
   http {
     bind-hostname = 0.0.0.0
+    bind-hostname = ${?MLEAP_SERVER_HOSTNAME}
     bind-port = 65327
+    bind-port = ${?MLEAP_SERVER_PORT}
   }
 }
--- a/mleap-serving/src/main/resources/application.conf
+++ b/mleap-serving/src/main/resources/application.conf
@@ -1,8 +1,6 @@
 ml.combust.mleap.serving {
   http {
-    hostname = 0.0.0.0
-    port = 65327
-    bind-hostname = ${ml.combust.mleap.serving.http.hostname}
-    bind-port = ${ml.combust.mleap.serving.http.port}
+    bind-hostname = 0.0.0.0
+    bind-port = 65327
   }
 }
--- a/mleap-serving/src/main/scala/ml/combust/mleap/serving/MleapConfig.scala
+++ b/mleap-serving/src/main/scala/ml/combust/mleap/serving/MleapConfig.scala
@@ -1,15 +1,16 @@
 package ml.combust.mleap.serving
 
 import com.typesafe.config.Config
+import scala.util.Properties
 
 /**
   * Created by hollinwilkins on 1/31/17.
   */
 case class HttpConfig(config: Config) {
   val hostname = config.getString("hostname")
-  val port = config.getInt("port")
+  val port = scala.util.Properties.envOrElse("MLEAP_SERVER_PORT", config.getInt("port"))
   val bindHostname = config.getString("bind-hostname")
-  val bindPort = config.getInt("bind-port")
+  val bindPort = scala.util.Properties.envOrElse("MLEAP_SERVER_PORT", config.getInt("bind-port"))
 }
 
 case class MleapConfig(config: Config) {
--- a/mleap-serving/src/main/scala/ml/combust/mleap/serving/MleapConfig.scala
+++ b/mleap-serving/src/main/scala/ml/combust/mleap/serving/MleapConfig.scala
@@ -1,14 +1,13 @@
 package ml.combust.mleap.serving
 
 import com.typesafe.config.Config
-import scala.util.Properties
 
 /**
   * Created by hollinwilkins on 1/31/17.
   */
 case class HttpConfig(config: Config) {
-  val bindHostname = scala.util.Properties.envOrElse("MLEAP_SERVER_HOSTNAME", config.getString("bind-hostname"))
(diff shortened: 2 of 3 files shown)


hollinwilkins:
@cappaberra This is great. I would like to note that we are going to be overhauling our server infrastructure at some point, when we merge this branch in: https://github.com/combust/mleap/tree/feature/mleap-grpc

We'll make sure to maintain this capability there as well.

mleap-serving/src/main/resources/application.conf
```diff
@@ -1,8 +1,6 @@
 ml.combust.mleap.serving {
   http {
-    hostname = 0.0.0.0
-    port = 65327
-    bind-hostname = ${ml.combust.mleap.serving.http.hostname}
-    bind-port = ${ml.combust.mleap.serving.http.port}
+    bind-hostname = 0.0.0.0
```
hollinwilkins: We can rewrite this line as:
```
bind-hostname = 0.0.0.0
basedir = ${?MLEAP_SERVER_HOSTNAME}
```

This will interpolate the environment variable

mleap-serving/src/main/resources/application.conf
```diff
@@ -1,8 +1,6 @@
...
+    bind-port = 65327
```
hollinwilkins: See comment on bind-hostname

mleap-serving/src/main/scala/ml/combust/mleap/serving/MleapConfig.scala
```diff
@@ -1,15 +1,14 @@
...
 
 /**
   * Created by hollinwilkins on 1/31/17.
   */
 case class HttpConfig(config: Config) {
-  val hostname = config.getString("hostname")
-  val port = config.getInt("port")
-  val bindHostname = config.getString("bind-hostname")
-  val bindPort = config.getInt("bind-port")
+  val bindHostname = scala.util.Properties.envOrElse("MLEAP_SERVER_HOSTNAME", config.getString("bind-hostname"))
```
hollinwilkins: See above comments, access the config value directly and rely on typesafe config to interpolate the environment variable.

mleap-serving/src/main/scala/ml/combust/mleap/serving/MleapConfig.scala
```diff
@@ -1,15 +1,14 @@
...
+  val bindPort = scala.util.Properties.envOrElse("MLEAP_SERVER_PORT", config.getString("bind-port")).toInt
```
hollinwilkins: See comment on bindHostname

hollinwilkins:
@cappaberra 

> Just realized.... prior to doing Http().bindAndHandle(), the variables (maybe?) should be validated for at least format, since they could have been passed in via the shell environment.

My take on this is that validation should be performed on external user inputs. Because the user here is the deployment team, it will error out pretty quickly if it's malformed and the deployment team isn't going to be malicious with it (hopefully).

mleap-serving/src/main/resources/application.conf
cappaberra: Ok! Done. Thank you, too, for the additional context on the future state of the MLeap server infrastructure.  We're looking forward to that, but this in the meantime will really help us!

davidecoccia:
@cappaberra thanks a lot for this, this is exactly what I was thinking about.
@hollinwilkins another approach to address the issue #325 would be to allow multiple models to coexist in the same instance and then allow the caller to go for the relevant endpoint e.g. http://instance1/model1/transform . But I still believe selecting the port gives more flexibility

hollinwilkins:
@davidecoccia Take a look at our new mleap-executor module and servers we are building. Will support many models from different sources: https://github.com/combust/mleap/tree/feature/mleap-grpc

cappaberra:
@hollinwilkins, to satisfy my IT Operations mentality, are you able to cut a minor release that incorporates this change so I can deploy a released artifact?  v0.10.1? Thanks!

hollinwilkins:
@cappaberra Just released 0.10.1

0xbliang:
> @cappaberra This is great. I would like to note that we are going to be overhauling our server infrastructure at some point, when we merge this branch in: https://github.com/combust/mleap/tree/feature/mleap-grpc
> We'll make sure to maintain this capability there as well.

Looks like we no longer have this capability after the overhaul? I tried both:
export MLEAP_SERVER_HOSTNAME=127.0.0.1
export MLEAP_HTTP_HOSTNAME=127.0.0.1

cappaberra:
I agree, @0xbliang... that functionality seems to have been lost.  That said, I don't think there's a need to have this as urgently since I was able to succesfully load 20 models into 1 MLeap 0.14.0 instance and not have any latency issues with 1 client / per model (20 total) hitting each model (you just need enough CPU cores!).  In the case that this is needed, which I'm sure there are some situations where it is / will be, @ancasarb, can you help putting this back?  I'll take a quick look, too.

0xbliang:
I figured out we can pass in `-Dserver.address=...`. So, it's no longer necessary.

cappaberra:
@0xbliang, oh, awesome! What about port?  Is it a similar convention?

0xbliang:
Yes, `-Dserver.port`

PR #368, opened 2018-05-22
fix pyspark init file

ancasarb:


cappaberra:
We ran into this as well... thank you for putting this into a pull request.

PR #382, opened 2018-07-06
remove argparse as a requirement for mleap scikit-learn

ancasarb:


python/mleap/version.py
```diff
@@ -16,4 +16,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.9.4"
+__version__ = "0.10.2-dev"
```
ancasarb: we should try to keep this in sync with the mleap version that we'll release

python/mleap/version.py
```diff
@@ -16,4 +16,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.9.4"
+__version__ = "0.11.0-dev"
```
ancasarb: we should try to keep this in sync with the mleap version that we release

PR #392, opened 2018-08-02
MLeap-391 Try to avoid calling Datafram.head to infer Vector's size

huafengw:


huafengw:
Hi @hollinwilkins , I realized that the default value in `getOrElse` is not lazy evaluated, which means `Datafram.head` will be called anyway. Really sorry for this buggy pull request...
Can you revert it or should I submit another one to fix this issue?

PR #398, opened 2018-08-10
Feature/databricks runtime

hollinwilkins:
This branch adds 3 submodules:

1. mleap-databricks-runtime-fat
2. mleap-databricks-runtime
3. mleap-databricks-runtime-testkit

# Mleap Databricks Runtime Fat

This module contains all of the MLeap dependencies required for execution in the Databricks ML Runtime. This will either exclude, mark provided or shade any 3rd party dependencies of MLeap training integrations and the MLeap runtime itself.

# Mleap Databricks Runtime

This is a facade module used for publishing to Maven. The POM file itself has no dependencies and the compiled JAR file is the contents of the `mleap-databricks-runtime-fat` assembly jar. This is the jar that should be used as a dependency where required.

# Mleap Databricks Runtime Testkit

A simple set of tests used to make sure that we can export model from Spark ML, TensorFlow and XGboost-Spark using the assembly jar. This is not published and most people will not ever need to worry about it.


hollinwilkins:
This PR has changes from https://github.com/combust/mleap/pull/401 included. Make sure to first merge https://github.com/combust/mleap/pull/401

Issue #380, opened 2018-07-03
Add support for Spark's OneHotEncoderEstimator/Model


lbustelo:
At some point in Spark 2.3.x there was a re-write of the OneHotEncoder support to better fit the Pipeline model. MLeap currently does not support these new types.

Beware... the docs for Spark say that there will be another renaming of this transformer in Spark 3.0

clocklear:
I'll take this issue.  It's pretty straightforward; sorry I overlooked this addition when I performed the upgrade to 2.3.

ancasarb:
Closing this as support for this was added in the PR merged above.

PR #408, opened 2018-08-22
[Ready to Merge] Support for Spark 2.3's OneHotEncoderModel

clocklear:
This PR addresses #380 and adds support for the new `OneHotEncoderEstimator`/`OneHotEncoderModel` that was introduced in Spark 2.3 (that I previously ignored because I thought it was a drop-in replacement for `OneHotEncoder`).  The biggest change is that 1HE now supports multiple inputs and outputs, so many things had to evolve to support that.  There are several changes in this PR, so I'll try to do my best to address them all.

* The `OneHotEncoderModel` inside `mleap-core` has been updated to support the new options that `OneHotEncoderModel` inside Spark supports (multiple inputs/outputs and a `HandleInvalid` param).
* Since the core model evolved, existing operations needed to be evolved as well.  `OneHotEncoder` and its associated ops (in both `mleap-runtime` and `mleap-spark`) have been updated to support the new model while maintaining backwards compatibility.
* New `OneHotEncoderV23` transformers and ops have been created to support the new `OneHotEncoderModel` in Spark.
* Since 1HE supports multiple outputs and we need a way to support that in our ops, `SimpleSparkOp` has been modified such that `sparkOutputs` can be any `Seq[ParamSpec]`.  This allows us to utilize `ArrayParamSpec` outputs instead of just `SimpleParamSpec`.  `SparkShapeSaver` and `SparkShapeLoader` have been modified accordingly to support this change as well.

All unit tests are passing.


mleap-runtime/src/main/scala/ml/combust/mleap/runtime/transformer/feature/OneHotEncoderV23.scala
```diff
@@ -0,0 +1,64 @@
...
+
+case class OneHotEncoderV23(override val uid: String =
+                              Transformer.uniqueName("one_hot_encoder_v23"),
+                            override val shape: NodeShape,
+                            override val model: OneHotEncoderModel)
+    extends MultiTransformer {
+  override val exec: UserDefinedFunction = {
+    // UGLY: I don't like this.  It's the best way forward from my current vantage point though.
+    // This stems from the fact that anonymous functions in scala cannot have varargs parameters.
+    shape.inputs.keys.size match {
```
hollinwilkins: Can return a Row(values: _*) for multiple outputs.

hollinwilkins: Take a look here for an example. Will be good too, because then we can support any number of outputs:

https://github.com/combust/mleap/blob/master/mleap-runtime/src/main/scala/ml/combust/mleap/runtime/transformer/classification/DecisionTreeClassifier.scala#L16-L42

clocklear: I'm already spreading the results of the model (see L23) as input to `Row`.  The problem is with not being able to take vararg inputs.

hollinwilkins: Ah, my bad. Here is example of vararg inputs:
https://github.com/combust/mleap/blob/master/mleap-runtime/src/main/scala/ml/combust/mleap/runtime/transformer/feature/VectorAssembler.scala#L25

bundle-ml/src/main/scala/ml/combust/bundle/dsl/Bundle.scala
```diff
@@ -58,6 +58,7 @@ object Bundle {
       val tokenizer = "tokenizer"
       val vector_assembler = "vector_assembler"
       val one_hot_encoder = "one_hot_encoder"
+      val one_hot_encoder_v23 = "one_hot_encoder_v23"
```
ancasarb: maybe we should call it "one_hot_encoder_estimator" for now, to match what Spark is calling it in 2.3?

ancasarb: I wonder whether we shouldn't keep the two models completely separate on MLeap side as well, allowing us not to have to modify the serialization (OneHotEncoderOp) on the existing one hot encoder. I'm just thinking that one hot encoding is very common and with the changes as they are now, people looking to upgrade to a newer version of MLeap in order to take advantage of bug fixes and new features available would need to retrain & re-serialize all their models. 

I know that we don't guarantee backwards compatibility yet, but maybe in this case, we don't really need to break the backwards compatibility? What do you think @clocklear @hollinwilkins?

clocklear: So, a bit of info to add some clarity.

The old 1HE is just called `OneHotEncoder`.  It is a stand-alone feature transformer.
The new 1HE has an estimator component called `OneHotEncoderEstimator`.  The output of the `fit()` method is a `OneHotEncoderModel`, which is the transformer that Mleap actually cares about.

I'm not strongly opinionated on the naming of the constant in `Bundle.scala`, other than I don't really think it should be named after the estimator.  After all, Mleap doesn't care about estimators or anything else that happens at train time.

hollinwilkins: Because we have 1 transformer and 1 core model for both v23 and previous versions, it would be best if we did not need to introduce an additional op name.

Default v23 to register "one_hot_encoder" as the V23 op, users can customize to the old one hot encoder if needed.

For below v23, default to the old one hot encoder.

hollinwilkins: A better long-term solution would be to support multiple ops per op name, but that is a bit out of scope for this PR.

ancasarb: @clocklear Sorry, I misread the changes in the op for the existing 1HE, makes sense now!

ancasarb: MLeap support for 2.3 is relatively new, so I think Hollin was suggesting for 2.3 to default to the new 1HE op (since the old one is deprecated and will be removed in 3.0 anyways), while keeping the old 1HE op for the previous versions.

clocklear: Ahh, thanks for the clarification @ancasarb!  That makes a lot of sense.

clocklear: So, if we don't register the old 1HE in 2.3, the parity tests no longer pass (because we didn't register an op for the transformer).  Is that ok?

I don't think we should do this until 1HE is actually removed (in 3.0).  It's deprecated now, but still fully supported, so I think we should still support it.  Again, I could be in the minority here.

ancasarb: I think you can try to create a single OneHotEncoderParitySpec class and then define the sparkTransformer something like this

```
  override val sparkTransformer: Transformer =
    if (org.apache.spark.SPARK_VERSION.equals("2.3.0")) {
// new 1HE pipeline
} else {
// old 1HE pipeline
}
```

clocklear: PR updated.

clocklear:
Travis failed build because these parity tests use 1HE and need to be updated.

```
org.apache.spark.ml.parity.regression.AFTSurvivalRegressionParitySpec
org.apache.spark.ml.parity.regression.GeneralizedLinearRegressionParitySpec
org.apache.spark.ml.parity.regression.LinearRegressionParitySpec
```

Will update these to take advantage of new 1HE.

PR #414, opened 2018-08-29
Feature/fix xgboost publish

hollinwilkins:
This will publish mleap-xgboost-spark to maven central.


hollinwilkins:
Already merged it looks like.

PR #415, opened 2018-08-29
Feature/bundle file system

hollinwilkins:
This PR provides support for saving/loading bundles using HDFS.

1. Introduces a new trait `BundleFileSystem` which can be used to implement non-local file systems for saving/loading bundles
2. Introduces a new module `bundle-hdfs` which provided an implementation of `BundleFileSystem` that integrates with HDFS
3. Introduces simpler interfaces for saving/loading bundle files using `java.net.URI` instead of `BundleFile`.

NOTES:

1. `BundleFile` interfaces still only support `jar` and `file` file systems, so if you want to use HDFS, need to use the new URI interfaces


bundle-hdfs/src/main/scala-2.10/HadoopBundleFileSystem.scala
```diff
@@ -0,0 +1,3 @@
+class HadoopBundleFileSystem {
```
ancasarb: I believe this can be removed.

PR #433, opened 2018-10-09
1HE NodeShape backwards compatibility

clocklear:
Relates to #417.

I'm embarrassed to say that my previous PR wasn't entirely complete.  We had some problem with our legacy models and it took some digging to figure out why -- the legacy serialized `NodeShape` object needs to be massaged a bit at load to work with the new implementation.   This PR adds that massaging.  Sorry I overlooked this the first time!


clocklear:
Oh it looks like some of the python stuff failed right at the beginning.

ancasarb:
@clocklear see https://github.com/combust/mleap/pull/431 on why python fails atm.

clocklear:
Alright, I'm going to pin scikit as well to make the build pass.  If we fix that in some other way, I can undo this.

PR #438, opened 2018-10-29
docker compose files

ganeshkrishnan1:
Makes it easier to start just by
docker-compose -f mleap.yml up -d


ancasarb:
This is a nice way of starting the docker container, but the volume path is per user environment, so not super generalizable. I think it would be ok to have this separate from the mleap repository.

Issue #464, opened 2018-12-22
Logistic Regression prediction probabilities


neilsummers:
Currently sklearn logistic classifier only supports predictions.

https://github.com/neilsummers/mleap/blob/master/python/mleap/sklearn/logistic.py
```
    def serialize_to_bundle(self, transformer, path, model_name):

        # compile tuples of model attributes to serialize
        attributes = list()
        attributes.append(('intercept', transformer.intercept_.tolist()[0]))
        attributes.append(('coefficients', transformer.coef_.tolist()[0]))
        attributes.append(('num_classes', 2)) # TODO: get number of classes from the transformer

        # define node inputs and outputs
        inputs = [{
                  "name": transformer.input_features,
                  "port": "features"
                }]

        outputs = [{
                  "name": transformer.prediction_column,
                  "port": "prediction"
                }]

        self.serialize(transformer, path, model_name, attributes, inputs, outputs)
```

Want to be able to access the probabilities from the classifier too.
This is already supported in the MLeap logistic model. Just need to add the appropriate outputs. I will create a PR with the change we made to access this for our own production model.

PR #465, opened 2018-12-22
sklearn Logistic Regression prediction probabilities (#464)

neilsummers:
Adding capability for access to prediction probabilities from sklearn logistic regression, per issue #464 .

Just copied across the outputs from https://github.com/combust/mleap/blob/master/python/mleap/sklearn/ensemble/forest.py serialize_to_bundle function in SimpleSerializer class.
The probabilities were already implemented in the scala version the mleap LR.

We have tested this in our production model with a test case which reproduced the probabilities from the sklearn predict_proba function and the probabilities from the scala function.


chris-russo:
@neilsummers is your team still using `mleap` in prod? I created an override for this before I saw your PR. I'd like to get this merged in as well.

neilsummers:
> @neilsummers is your team still using `mleap` in prod? I created an override for this before I saw your PR. I'd like to get this merged in as well.

Yes, we are using mleap sklearn in prod. We are having to still use our workaround (almost a year later) until this is merged. Would like to see this merged too.

cc: @ancasarb @seme0021 @hollinwilkins

python/mleap/sklearn/logistic.py
```diff
@@ -75,12 +75,23 @@ def serialize_to_bundle(self, transformer, path, model_name):
...
+
+        outputs.append({
+            "name": "raw_prediction",
+            "port": "raw_prediction"
+        })
+
+        outputs.append({
+            "name": "probability",
+            "port": "probability"
+        })
```
ancasarb: Just wanted to make sure, this isn't going to generate the raw predictions/probabilities in sklearn, it's just going to serialize a bundle that at scoring time will generate those. Is that the intention of this change?

python/mleap/sklearn/logistic.py
```diff
@@ -75,12 +75,23 @@ def serialize_to_bundle(self, transformer, path, model_name):
...
+            "port": "prediction"
+        })
+
+        outputs.append({
+            "name": "raw_prediction",
+            "port": "raw_prediction"
+        })
+
+        outputs.append({
+            "name": "probability",
```
ancasarb: Could we make the name of the raw_prediction and probability columns configurable in the mleap_init function? Also maybe allow for these to be null as default, in which case they wouldn't be added to the outputs list. What do you think?

python/mleap/sklearn/logistic.py
chris-russo: Hi @ancasarb – that's correct. This just tell's the loaded JVM MLeap bundle to produce predicted and raw probabilities at scoring time.

python/mleap/sklearn/logistic.py
chris-russo: What about `transformer.prediction_column} + "_raw"` and `transformer.prediction_column + "_proba"`for the `name` keys of these outputs, respectively?

ancasarb: I would like to still keep the option of generating these additional outputs only when specified to do, it would be similar to how Spark handles this as well.

python/mleap/sklearn/logistic.py
ancasarb: Makes sense, thanks!

python/mleap/sklearn/logistic.py
ancasarb: Let me know if you need any help to make these changes, happy to help.

ancasarb:
hey @neilsummers, i'll close this PR for now, please reopen it if you're still looking to get it merged. thank you!

PR #469, opened 2019-01-08
Add health check to mleap-serving

Khiem-Tran:
### Purpose
Adding health check endpoint from spring-actuator to mleap-serving

### Approach
- Adding spring-actuator
- Update API document

### DIFF
--- a/mleap-spring-boot/README.md
+++ b/mleap-spring-boot/README.md
@@ -73,7 +73,7 @@ curl --header "Content-Type: application/json" \
 
 Format can be either `ml.combust.mleap.binary` or `ml.combust.mleap.json` and your leap frame needs to be encoded with that format.
 
-Note: The above endpoints are available either using: 
+Note: The above endpoints are available either using:
 - JSON (`Content-Type` header set to `application/json`)
 - Protobuf (`Content-Type` header set to `application/x-protobuf`).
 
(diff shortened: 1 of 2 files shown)


mleap-spring-boot/src/main/resources/mleap_serving_1.0.0_swagger.yaml
```diff
@@ -189,12 +189,28 @@ paths:
...
+      responses:
+        '200':
+          description: successful operation
+          content:
+            application/json:
+              schema:
+               $ref: '#/components/schemas/HealthCheckResponse'
+            application/x-protobuf:
+              schema:
+                $ref: '#/components/schemas/HealthCheckResponse'
```
ancasarb: There's a couple of endpoints that are enabled by default by the actuator.

```
{
    "_links": {
        "self": {
            "href": "http://localhost:8080/actuator",
            "templated": false
        },
        "health": {
            "href": "http://localhost:8080/actuator/health",
            "templated": false
        },
        "info": {
            "href": "http://localhost:8080/actuator/info",
            "templated": false
        }
    }
}
```
Instead of adding them to the swagger yaml file, could you please just add a note in the README.md file in this project, that Spring Actuator endpoints enabled by default are also available? Thank you!

PR #476, opened 2019-01-16
Lower sklearn requirement to 0.17

mdgoldberg:
I am currently limited to using scikit-learn 0.17 due to the fact that model persistence isn't guaranteed to work when models were trained/written to a file in one version and then loaded from another version (https://scikit-learn.org/stable/modules/model_persistence.html#security-maintainability-limitations).

I don't see any reason mleap requires scikit-learn >= 0.18.dev0 as opposed to 0.17, so would it be possible to relax that constraint slightly?

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -42,7 +42,7 @@ REQUIRED_PACKAGES = [
       'six >= 1.10.0',
       'scipy>=0.13.0b1',
       'pandas>=0.18.1',
-      'scikit-learn>=0.18.dev0',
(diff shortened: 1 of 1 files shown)


mdgoldberg:
@ancasarb @hollinwilkins ? Sorry wasn't sure who best to tag.

ancasarb:
0.17 is quite old (released around 2015), in fact, I'd like us to try to keep updating the dependencies so that we use the newer/latest versions.

PR #493, opened 2019-02-26
Adding support for Spark 2.4

ancasarb:
Only difference was that OneVsRestModel in Spark now outputs raw_prediction(s) too, so I've updated the mleap model & transformer to do the same.


README.md
```diff
@@ -49,6 +49,7 @@ Choose the right verison of the `mleap-spark` module to export your pipeline. Th
 
 | MLeap Version | Spark Version |
 |---------------|---------------|
+| 0.14.0        | 2.3           |
```
hollinwilkins: Shouldn't this be spark version 2.4?

ancasarb: updated.

ancasarb:
closing this, as https://github.com/combust/mleap/pull/496 contains changes for both spark 2.4 and scala 2.12

Issue #504, opened 2019-03-31
Spark schema to Mleap schema


gitabu:
When we need to convert basic spark schema to mleap schema, we have to provide the spark dataset. 

To provide dataset, we have create spark data frame by using SparkContext. I want to support spark schema with basic data types so that when we convert spark model to mleap in production, one dont need to convert the spark schema to mleap schema. 

Currently I am doing this in my project but want to contribute it to mleap.

This feature will help programmer with minimum code changes when he developed model in spark mllib.

ancasarb:
Hey @gitabu, we already have a way of converting from Spark schema to MLeap schema, you don't have to do this manually.

```
import org.apache.spark.sql.mleap.TypeConverters
TypeConverters.sparkSchemaToMleapSchema(dataset).print()
```

Hope this helps!

ancasarb:
Closing this, please re-open if you find that the above solution doesn't quite meet your needs and happy to discuss this further! Thank you!

gitabu:
I am looking for function where we can convert spark schema to mleap schema without providing dataset. Similar to below util:
 
def mleapSchemaToSparkSchema(schema: types.StructType): StructType = {
    val fields = schema.fields.map(mleapFieldToSparkField)
    StructType(fields)
  }

ancasarb:
How or where are you planning to use the mleap schema? Is it for scoring?

bhrigs:
Hi @ancasarb 
Can you help me with the python version of the library import - `import org.apache.spark.sql.mleap.TypeConverters` ?
I need to use it to predict on sagemaker endpoint.

PR #507, opened 2019-04-04
#504 Spark schema to Mleap schema

gitabu:
When we need to convert basic spark schema to mleap schema, we have to provide the spark dataset.

To provide dataset, we have create spark data frame by using SparkContext. 

This change will support spark schema with basic data types so that when user convert spark model to mleap in code, s/he doesn't need to convert the spark schema to mleap schema manually.  By using the utility function he can convert the spark schema to Mleap schema and pass it to DefaultLeafFrame.

Please note that this for basic data types any custom data set is not supported for now.


ancasarb:
Hey @gitabu, we already have a way of converting from Spark schema to MLeap schema, you don't have to do this manually.

```import org.apache.spark.sql.mleap.TypeConverters
TypeConverters.sparkSchemaToMleapSchema(dataset).print()```

Hope this helps!

ancasarb:
Closing this, please let me know if there's anything missing from the existing ```org.apache.spark.sql.mleap.TypeConverters```, seems to work with all data types.

PR #538, opened 2019-07-20
Issues/525 mleap-spring-boot load model

danielpoonwj:
### Purpose

Allow loading models at startup instead of needing to send a HTTP request after the spring boot process has started. This would especially be relevant for docker-based deployments.

#### Misc

- Change Docker base image from `openjdk:8` to `openjdk:8-jre-slim` to reduce image size. Only JRE instead of full JDK is needed for runtime.

### Approach

Implement `ModelLoader` Component with a PostConstruct method to load one or many models from config files. Each config file should take the form of a JSON file with the exact same structure as the POST body you would typically send to the `/models` endpoint to load the model.

This functionality can be activated by passing the configuration property `mleap.model.config`. Specifying a path to a single config will load that single config, while specifying a path to a directory will load all files from that directory.

### Example Docker workflow

The project directory is assumed to look like this:

```
├── Dockerfile
├── config
│   ├── application.properties
│   └── startup
│       ├── airbnb-lr.json
│       └── airbnb-rf.json
└── models
    ├── airbnb.model.lr.zip
    └── airbnb.model.rf.zip
```

After building the docker image and starting the container, you can verify both models are loaded by querying their respective `/meta` endpoints.

#### config/startup/airbnb-lr.json

```json
{
    "config": {
        "diskTimeout": 525600000,
        "memoryTimeout": 525600000
    },
    "modelName": "airbnb-lr",
    "uri": "file:/opt/docker/models/airbnb.model.lr.zip"
}
```

#### config/startup/airbnb-lr.json

```json
{
    "config": {
        "diskTimeout": 525600000,
        "memoryTimeout": 525600000
    },
    "modelName": "airbnb-rf",
    "uri": "file:/opt/docker/models/airbnb.model.rf.zip"
}
```

#### config/application.properties

```
mleap.model.config=/opt/docker/config/startup
```

#### Dockerfile

```Dockerfile
FROM combustml/mleap-spring-boot

# Override default server port if needed
ENV SERVER_PORT=8080
EXPOSE 8080

# Copy models
COPY models/*.zip models/

# Copy application.properties
COPY config/application.properties ./application.properties

# Copy startup configs
COPY config/startup/*.json ./config/startup/
```


mleap-spring-boot/src/main/scala/ml/combust/mleap/springboot/ModelLoader.scala
```diff
@@ -0,0 +1,52 @@
...
+import ml.combust.mleap.pb
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+
+import scala.collection.JavaConverters._
+import java.nio.file.{Paths, Files, Path}
+import scalapb.json4s.Parser
+
+@Component
+class ModelLoader {
```
ancasarb: We should autowire the mleap executor and parser that get created in the StarterConfiguration and use them in this class 

```
class ModelLoader(@Autowired val mleapExecutor: MleapExecutor,
                  @Autowired val jsonParser: Parser)
```

instead of creating a new parser
```
private val jsonParser = new Parser()
```

and doing

```
StarterConfiguration.getMleapExecutor
```

I believe the getMleapExecutor() is then not needed?

PR #567, opened 2019-09-14
reorganize python code, separate tests out 

lydian:
The pull request is aiming on simplifying the python development by using popular tools/structure in python world :) 

- move out all the unit test code under tests/ folder, so that we don't mix the test code with actual code. 
-  added tox.ini so that it is much more easier to test with different python version. This probably help debug on python 3 support PR:  https://github.com/combust/mleap/pull/565 
- the test script will show the coverage as well  
- travis CI is not naturally support tox, therefore update the script to run with different python. (Note that I didn't enable py3 yet, because the current code will fail to build for py3)
- I originally plan to follow pep8 (python style guide, but that makes the entire CR really huge, given this is my first pull request, I'd  like to keep it simple, and maybe try doing that in the follow up pull request instead.


ancasarb:
@lydian thanks for the PR, this is great! Could you please take a look at why the build failed? thanks again!

lydian:
Sure! 
It is actually flaky tests, which I had the same error on the master branch. 
The main reason is that python dict doesn't guarantee the output order. To resolve the issue, I replace the dict with OrderedDict :) 

For more detail, you can find it here: 
https://stackoverflow.com/questions/5629023/the-order-of-keys-in-dictionaries

python/mleap/sklearn/preprocessing/data.py
```diff
@@ -1097,8 +1098,12 @@ def fit_transform(self, X, y=None, **fit_params):
     def serialize_to_bundle(self, path, model_name):
         # compile tuples of model attributes to serialize
         attributes = list()
-        attributes.append(("labels", self.labels.keys()))
-        attributes.append(("values", Vector(self.labels.values())))
+        labels = (
+            sorted(self.labels.keys())
+            if not isinstance(self.labels, OrderedDict) else self.labels.keys()
```
lucagiovagnoli: Does it make sense to save the self.labels as an OrderedDict once and for all ? Then we don't need to make sure we always sort or convert to OrderedDict everywhere else. Basically create one on line 1081

python/mleap/sklearn/preprocessing/data.py
```diff
@@ -1097,8 +1098,12 @@ def fit_transform(self, X, y=None, **fit_params):
...
+        )
+        attributes.append(("labels", labels))
+        attributes.append(("values", Vector([self.labels[label] for label in labels])))
```
lucagiovagnoli: we could just do self.labels.values() if we had self.labels: OrderedDict here

.travis.yml
```diff
@@ -13,10 +13,9 @@ python:
   - "2.7"
 
 install:
-  pip install --user -r ./python/requirements.txt
-
+    -   pip install --user tox
 script:
-  - nosetests ./python -v --exclude-dir=./python/mleap/pyspark
+  - tox -c python/tox.ini -e $(echo py$TRAVIS_PYTHON_VERSION | tr -d .) -v
```
lucagiovagnoli: isn't `tox -c python/tox.ini` sufficient? All version in envlist from the tox.ini will be run.
eg:
```
envlist = py27,py36
```

lydian: the .travis is not working well with tox system. The workaround is to make travis install 2 different version of python and make tox run for each specific python version. 

I am using a solution mentioned here: https://stackoverflow.com/a/42309035

Note that there's an plugin https://github.com/tox-dev/tox-travis which will resolve the issue. However, the package assumes your tox.ini is on the root folder, and I am not sure if we want to move tox.ini out of python folder, because I make working folder to be `python`.   Also the old `run_test` script also seems to assume python is the working folder when running the tests, therefore I am not using this approach and just update the travis command to run tox instead.

lucagiovagnoli: ok, makes sense

lucagiovagnoli:
My PR https://github.com/combust/mleap/pull/569 passes the build, and it's rebased on top of this. So I think your Scala failure is a flake. It'd be good to ticket this flake, but you could try re-run the tests in the meantime? Also if it's a scala failure, it's unlikely your code cause that

python/mleap/sklearn/preprocessing/data.py
lydian: sure, I'll update that :)

python/tox.ini
```diff
@@ -0,0 +1,21 @@
+[tox]
+envlist = py27
+skipdist = true
+
+[testenv]
+deps =
+        -rrequirements.txt
```
ancasarb: just wanted to make sure this is intended and not a typo?

lucagiovagnoli: Oh yes, that could look like a typo! but it's not :)

When checking the options for pip install we get:
```
$ pip install --help
[...]
  -r, --requirement <file>    Install from the given requirements file. This option can be used multiple times.
```

And the space between -r and requirements.txt is not required.

juhoautio:
@lydian maybe you could help: when I run `make test`, I get this:

```
  py27: commands succeeded
ERROR:  py36: InterpreterNotFound: python3.6
  py37: commands succeeded
```

I thought tox would pick whatever 3.6.x I have installed with pyenv, but it doesn't seem to find it?

I have:

```
pyenv versions
  system
  2.6.9
* 2.7.15 (set by /Users/JuhoAutio/.pyenv/version)
  3.6.8
  3.7.0
* 3.7.2 (set by /Users/JuhoAutio/.pyenv/version)
```

And:

```
pyenv version
2.7.15 (set by /Users/JuhoAutio/.pyenv/version)
3.7.2 (set by /Users/JuhoAutio/.pyenv/version)
```

I also tried running `make env` once more, but still getting `ERROR: InterpreterNotFound: python3.6`.

PR #568, opened 2019-09-16
pyspark wrapper for StringMap

juhoautio:
Add a pyspark wrapper for the StringMap transformer (which is a custom transformer by mleap).

It seems like there are no previous examples of creating a pyspark wrapper for a custom mleap transformer. This may act as one?

If this should be unit-tested, is there any guidance for how that would be done?

----

I found that a `feature.py` previously existed:
https://github.com/combust/mleap/pull/327/files#diff-22934a469cf7ba2c623e2bc819ed776f
..but it was deleted because mleap's custom OneHotEncoder was removed in favor of the one provided by Spark.

This code is somewhat based on that pre-existing `feature.py`.


juhoautio:
The first attempt on Travis failed with this (most likely unrelated) error:

[info] *** 1 TEST FAILED ***
4625[error] Failed tests:
4626[error] 	ml.combust.mleap.grpc.server.GrpcSpec
4627[info] Run completed in 3 minutes, 40 seconds.
4628[info] Total number of tests run: 0
4629[info] Suites: completed 0, aborted 0
4630[info] Tests: succeeded 0, failed 0, canceled 0, ignored 0, pending 0
4631[info] No tests were executed.
4632[error] (mleap-grpc-server/test:test) sbt.TestsFailedException: Tests unsuccessful
4633[error] Total time: 431 s, completed Sep 16, 2019 9:13:40 PM

Is that test known to be flaky? Should it be disabled at least for now?

I found another recent PR build that had failed with the same error:
https://travis-ci.org/combust/mleap/builds/585321081

Ran tests locally with `sbt clean compile test` -> no errors.

I didn't know how else to trigger a new build on Travis, so I updated the commit timestamp & force-pushed here.

mleap-grpc-server/src/test/scala/ml/combust/mleap/grpc/server/GrpcSpec.scala
```diff
@@ -6,12 +6,14 @@ import akka.testkit.TestKit
...
 import ml.combust.mleap.executor.testkit.TransformServiceSpec
-import org.scalatest.{BeforeAndAfterAll, BeforeAndAfterEach}
+import org.scalatest.{BeforeAndAfterAll, BeforeAndAfterEach, Ignore}
 import org.scalatest.concurrent.ScalaFutures
 
 import scala.concurrent.duration._
 import ml.combust.mleap.grpc.server.TestUtil._
 
+// Ignored temporarily because of random failures on Travis
+@Ignore
```
ancasarb: I believe the flaky one is the "throttling" in TransformServiceSpec, could you please try ignoring just that for now? thank you!

python/mleap/pyspark/feature.py
```diff
@@ -0,0 +1,38 @@
...
+        assert handleInvalid in ['error', 'keep'], 'Invalid value for handleInvalid: {}'.format(handleInvalid)
+        super(StringMap, self).__init__()
+        labels_scala_map = _jvm() \
+            .scala \
+            .collection \
+            .JavaConverters \
+            .mapAsScalaMapConverter(labels) \
+            .asScala() \
+            .toMap(_jvm().scala.Predef.conforms())
+        handle_invalid_jvm = _jvm().ml.combust.mleap.core.feature.HandleInvalid.__getattr__(
```
ancasarb: I think this needs updating to StringMapHandleInvalid.

python/mleap/pyspark/feature.py
```diff
@@ -0,0 +1,38 @@
...
+            handleInvalid.capitalize() + '$').__getattr__('MODULE$')
+        string_map_model = self._new_java_obj("ml.combust.mleap.core.feature.StringMapModel",
+                                              labels_scala_map, handle_invalid_jvm, defaultValue)
+        self._java_obj = self._new_java_obj("org.apache.spark.ml.mleap.feature.StringMap", self.uid, string_map_model)
+        self.setInputCol(inputCol)
+        self.setOutputCol(outputCol)
+
+    @classmethod
+    def from_dataframe(cls, df, key_col, value_col):
```
ancasarb: when/how is this method going to be used/needed?

python/mleap/pyspark/feature.py
juhoautio: True.

How about this:

> If this should be unit-tested, is there any guidance for how that would be done?

(and I think it should be unit-tested :))

ancasarb: I agree about the unit tests, I am not sure we had any for the previous one hot encoder though. Can you take a look at how that could work?

python/mleap/pyspark/feature.py
```diff
@@ -0,0 +1,38 @@
+from pyspark.ml.util import JavaMLReadable, JavaMLWritable, _jvm
+from pyspark.ml.wrapper import JavaTransformer
+from pyspark.ml.param.shared import HasInputCol, HasOutputCol
+
+
+class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, JavaMLWritable):
```
ancasarb: maybe this should be in a string_map.py file, as we're likely to add support for some more transformers?

python/mleap/pyspark/feature.py
juhoautio: It has been used to convert a regular spark DataFrame into a a string transformer. I think it's a common use case. For example:
- Prepare some mappings with spark. eg. most simple case: `df = spark.table('some.table')`
- Converting the result into a python dict is not _that_ trivial, so having a helper for it makes sense.
- It allows keeping the code a bit cleaner, ie. this one-liner:
	```python
	string_map = StringMap.from_dataframe(df, 'key_col', 'value_col')`
	```
  instead of this:
	```python
	rows = df.select(['key_col', 'value_col']).collect()
	mappings_dict = {row[0]: float(row[1]) for row in rows}
	string_map = StringMap(mappings_dict, 'key_col', 'value_col')
	```

However, maybe the API could be different? Another option would be to check the type of `labels` in `__init__`, and if it's a `DataFrame`, collect it into a dict. How would you like that?

mleap-grpc-server/src/test/scala/ml/combust/mleap/grpc/server/GrpcSpec.scala
juhoautio: Changed to ignore "throttling" only

python/mleap/pyspark/feature.py
ancasarb: I think it's okay as is, I just wanted to understand a bit better its use :)

ancasarb:
looks good, thank you!

python/mleap/pyspark/feature.py
juhoautio: WIP pyspark unit testing: https://github.com/combust/mleap/pull/571

Issue #570, opened 2019-09-24
Add support for spark-extensions transformers in Pyspark


ancasarb:
Creating this parent issue to track progress on adding support for spark-extensions transformers in Pyspark.

- [x] Support for StringMap 
- [ ] Support for MathBinary
- [ ] Support for MathUnary
- [ ] Support for WordLengthFilter
- [ ] Support for MultinomialLabeler
- [ ] Support for Imputer
- [ ] Support for OneVsRest
- [ ] Support for SupportVectorMachine
- [ ] Add unit test support for Pyspark transformers

Related issues:
- https://github.com/combust/mleap/issues/551
- https://github.com/combust/mleap/issues/552
- https://github.com/combust/mleap/issues/477
- https://github.com/combust/mleap/issues/495

mtsol:
how long will it take?

juhoautio:
>  Add unit test support for Pyspark transformers

I have added most of the building blocks for that here: https://github.com/combust/mleap/pull/571

lucagiovagnoli:
MathUnary implementation at: https://github.com/combust/mleap/pull/666

lucagiovagnoli:
MathBinary implementation: https://github.com/combust/mleap/pull/672

RuxuePeng:
can we also add VectorSizeHint (A feature transformer)?
my model has a stage for VectorSizeHint and it failed during mleap serialization.

jsleight:
> can we also add VectorSizeHint (A feature transformer)?

Seems reasonable.  @RuxuePeng do you have any interest in submitting a PR?  I'm happy to help you review it and merge 😄

johnnyasd12:
Hi
Is there a better way to serialize a custom transformer with python now?
According to https://github.com/combust/mleap-docs/issues/15 it's hard to do it when someone is unfamiliar with Scala.
Is there any update?
Thanks.

LucasSchelkes-BA:
@jsleight thanks for giving insight into this topic! For our coperate data science project we also need custom transformers in pyspark for mleap. Are there any updates on this since 2021? Its would definetly be a very welcomed feature!

jsleight:
Don't think there's been any updates w.r.t. simplifying the process of making custom transformers.

PR #569, opened 2019-09-18
Tests running on py36, py37

lucagiovagnoli:
**!NOTE!** Based on @lydian's PR #567 
Fixes #566 

**CHANGELOG**:
- Running tests for py36 and py37
- Stricter requirements on pandas, because pandas>=0.25 breaks tests:
```
DataFrame.columns classes are not equivalent
[left]:  Index(['a'], dtype='object')
[right]: RangeIndex(start=0, stop=1, step=1)
```

PR #571, opened 2019-09-24
pyspark unit test tooling + test StringMap pyspark wrapper

juhoautio:
1. This implements the task `Add unit test support for Pyspark transformers` of https://github.com/combust/mleap/issues/570.

2. Validation & API improvements to string_map pyspark wrapper
- Validate the args as much as possible before passing to JVM, because otherwise user gets some hard to interpret Py4j error
- Use the names familiar from scala for args

----

Flipped the build order in `travis.sh` to run sbt first because pyspark tests require the compiled classes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ python:
 install:
     -   pip install --user tox
 script:
-  - tox -c python/tox.ini -e $(echo py$TRAVIS_PYTHON_VERSION | tr -d .) -v
   - travis/travis.sh
+  - tox -c python/tox.ini -e $(echo py$TRAVIS_PYTHON_VERSION | tr -d .) -v
 
 notifications:
   on_success: change
--- a/mleap-spark-extension/build.sbt
+++ b/mleap-spark-extension/build.sbt
@@ -5,15 +5,22 @@ Dependencies.sparkExtension
 
 val writeRuntimeClasspathToFile = taskKey[Unit]("Writes runtime classpath to a file")
 
-writeRuntimeClasspathToFile := {
+// TODO This produces the same output on command line
+//
+//  sbt --error "export mleap-spark-extension/runtime:fullClasspath" > fullClasspath.txt
+//
+// (--error is needed to disable info logging printed into the the file as well)
+//
+// TODO I couldn't find how to run the same "command" as scala from code here, but I guess it should be doable.
+//  That would allow simplifying the "impl" of writeRuntimeClasspathToFile task below.
+
+// this works in all cases: especially also if sources are compiled due to `sbt test` this task is run
+writeRuntimeClasspathToFile <<= Def.task {
+  // sbt complains about deprecated syntax, but unfortunately := can't be used with triggeredBy with sbt 0.13.
+  // Fix for triggeredBy is only available in 0.13.14, see https://github.com/sbt/sbt/issues/1444
   val file = reflect.io.File("mleap-spark-extension/target/classpath-runtime.txt")
   println(f"writeRuntimeClasspathToFile -> ${file.toAbsolute}")
   file
     .writeAll(Path.makeString((dependencyClasspath in Runtime)
       .value.map(_.data)))
-}
-
-// TODO this didn't work exactly as intended.. requires explicitly calling 'sbt compile'. eg. 'sbt test' skips this
-//  how to make sbt always run this always, for example as part of 'sbt test'?
-// run writeRuntimeClasspathToFile as part of 'compile' of this module
-//compile := {(compile in Compile) dependsOn writeRuntimeClasspathToFile}.value
+}.triggeredBy(compile in Compile)
--- a/mleap-spark-extension/build.sbt
+++ b/mleap-spark-extension/build.sbt
@@ -2,3 +2,16 @@ import ml.combust.mleap.{Dependencies, Common}
 
 Common.defaultMleapSettings
 Dependencies.sparkExtension
+
+val writeRuntimeClasspathToFile = taskKey[Unit]("Writes runtime classpath to a file")
+
+writeRuntimeClasspathToFile := {
+  val file = reflect.io.File("mleap-spark-extension/target/classpath-runtime.txt")
+  println(f"writeRuntimeClasspathToFile -> ${file.toAbsolute}")
+  file
+    .writeAll(Path.makeString((dependencyClasspath in Runtime)
+      .value.map(_.data)))
+}
+
+// run writeRuntimeClasspathToFile as part of 'compile' of this module
+compile := {(compile in Compile) dependsOn writeRuntimeClasspathToFile}.value
--- a/mleap-spark-extension/build.sbt
+++ b/mleap-spark-extension/build.sbt
@@ -13,5 +13,7 @@ writeRuntimeClasspathToFile := {
       .value.map(_.data)))
 }
 
+// TODO this didn't work exactly as intended.. requires explicitly calling 'sbt compile'. eg. 'sbt test' skips this
+//  how to make sbt always run this always, for example as part of 'sbt test'?
 // run writeRuntimeClasspathToFile as part of 'compile' of this module
-compile := {(compile in Compile) dependsOn writeRuntimeClasspathToFile}.value
+//compile := {(compile in Compile) dependsOn writeRuntimeClasspathToFile}.value
--- a/mleap-spark-extension/build.sbt
+++ b/mleap-spark-extension/build.sbt
@@ -5,23 +5,16 @@ Dependencies.sparkExtension
 
 val writeRuntimeClasspathToFile = taskKey[Unit]("Writes runtime classpath to a file")
 
-// TODO This produces the same output on command line
-//
-//  sbt --error "export mleap-spark-extension/runtime:fullClasspath" > fullClasspath.txt
-//
-// (--error is needed to disable info logging printed into the the file as well)
-//
-// TODO I couldn't find how to run the same "command" as scala from code here, but I guess it should be doable.
-//  That would allow simplifying the "impl" of writeRuntimeClasspathToFile task below.
-
 // this works in all cases: especially also if sources are compiled due to `sbt test` this task is run
 writeRuntimeClasspathToFile <<= Def.task {
   // sbt complains about deprecated syntax, but unfortunately := can't be used with triggeredBy with sbt 0.13.
   // Fix for triggeredBy is only available in 0.13.14, see https://github.com/sbt/sbt/issues/1444
-  val file = reflect.io.File("mleap-spark-extension/target/classpath-runtime.txt")
+
+  val file = reflect.io.File(f"mleap-spark-extension/target/classpath-runtime_${(scalaVersion in ThisBuild).value}.txt")
   println(f"writeRuntimeClasspathToFile -> ${file.toAbsolute}")
-  val classPathString = Path.makeString((dependencyClasspath in Runtime)
-    .value.map(_.data))
+
+  val classPathString = Path.makeString((dependencyClasspath in Runtime).value.map(_.data))
   println(f"classPathString: ${classPathString}")
+
   file.writeAll(classPathString)
 }.triggeredBy(compile in Compile)
--- a/mleap-spark-extension/build.sbt
+++ b/mleap-spark-extension/build.sbt
@@ -20,7 +20,8 @@ writeRuntimeClasspathToFile <<= Def.task {
   // Fix for triggeredBy is only available in 0.13.14, see https://github.com/sbt/sbt/issues/1444
   val file = reflect.io.File("mleap-spark-extension/target/classpath-runtime.txt")
   println(f"writeRuntimeClasspathToFile -> ${file.toAbsolute}")
-  file
-    .writeAll(Path.makeString((dependencyClasspath in Runtime)
-      .value.map(_.data)))
+  val classPathString = Path.makeString((dependencyClasspath in Runtime)
+    .value.map(_.data))
+  println(f"classPathString: ${classPathString}")
+  file.writeAll(classPathString)
 }.triggeredBy(compile in Compile)
--- a/python/mleap/pyspark/__init__.py
+++ b/python/mleap/pyspark/__init__.py
@@ -1,5 +1,13 @@
+import pyspark.ml
 import mleap.pyspark
+import mleap.pyspark.feature
+from mleap.pyspark.feature.string_map import StringMap
 import sys
 
 sys.modules['pyspark.ml.mleap'] = mleap
 sys.modules['pyspark.ml.mleap.pyspark'] = sys.modules['mleap.pyspark']
+sys.modules['pyspark.ml.mleap.feature'] = sys.modules['mleap.pyspark.feature']
+
+sys.modules['pyspark.ml'].mleap = mleap
+sys.modules['pyspark.ml'].mleap.feature = sys.modules['mleap.pyspark.feature']
+sys.modules['pyspark.ml'].mleap.feature.StringMap = StringMap
--- a/python/mleap/pyspark/feature/string_map.py
+++ b/python/mleap/pyspark/feature/string_map.py
@@ -9,7 +9,7 @@ class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, Java
     def __init__(self, labels, inputCol=None, outputCol=None, handleInvalid='error', defaultValue=0.0):
         """
         __init__(self, labels, inputCol=None, outputCol=None, handleInvalid='error', defaultValue=0.0)
-        labels is a dict {string: double}
+        labels must be a dict {string: double} or a spark DataFrame with columns inputCol & outputCol
         handleInvalid: how to handle missing labels: 'error' (throw an error), or 'keep' (map to the default value)
         """
         super(StringMap, self).__init__()
--- a/python/mleap/pyspark/feature/string_map.py
+++ b/python/mleap/pyspark/feature/string_map.py
@@ -8,23 +8,22 @@ from pyspark.sql import DataFrame
 class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, JavaMLWritable):
     def __init__(self, labels, inputCol=None, outputCol=None, handleInvalid='error', defaultValue=0.0):
         """
-        __init__(self, labels, inputCol=None, outputCol=None, handleInvalid='error', defaultValue=0.0)
+        :param labels: a dict {string: double}
+        :param handleInvalid: how to handle missing labels: 'error' (throw), or 'keep' (map to defaultValue)
+        :param defaultValue: value to use if key is not found in labels
+        """
+        """
         labels must be a dict {string: double} or a spark DataFrame with columns inputCol & outputCol
-        handleInvalid: how to handle missing labels: 'error' (throw an error), or 'keep' (map to the default value)
+        handleInvalid: 
         """
         super(StringMap, self).__init__()
-
-        if isinstance(labels, DataFrame):
-            assert inputCol, 'inputCol is required when labels is a DataFrame'
-            assert outputCol, 'outputCol is required when labels is a DataFrame'
-            labels = {row[0]: float(row[1]) for row in labels.select([inputCol, outputCol]).collect()}
 
         def validate_args():
             """
             validate args early to avoid failing at Py4j with some hard to interpret error message
             """
             assert handleInvalid in ['error', 'keep'], 'Invalid value for handleInvalid: {}'.format(handleInvalid)
-            assert isinstance(labels, dict), 'labels must be a dict or a DataFrame, got: {}'.format(type(labels))
+            assert isinstance(labels, dict), 'labels must be a dict, got: {}'.format(type(labels))
             for (key, value) in labels.items():
                 assert isinstance(key, six.string_types), \
                     'label keys must be a string type, got: {}'.format(type(key))
@@ -46,3 +45,15 @@ class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, Java
         self._java_obj = self._new_java_obj("org.apache.spark.ml.mleap.feature.StringMap", self.uid, string_map_model)
         self.setInputCol(inputCol)
         self.setOutputCol(outputCol)
+
+    @classmethod
+    def from_dataframe(cls, labels_df, inputCol, outputCol, handleInvalid='error', defaultValue=0.0):
+        """
+        :param labels_df: a spark DataFrame whose columns include inputCol:string & outputCol:double.
+        See StringMap() for other params.
+        """
+        assert isinstance(labels_df, DataFrame), 'labels must be a DataFrame, got: {}'.format(type(labels_df))
+        labels_dict = {row[0]: float(row[1]) for row in
+                       labels_df.select([inputCol, outputCol]).collect()}
+        return cls(labels_dict, inputCol=inputCol, outputCol=outputCol, handleInvalid=handleInvalid,
+                   defaultValue=defaultValue)
--- a/python/mleap/pyspark/feature/string_map.py
+++ b/python/mleap/pyspark/feature/string_map.py
@@ -1,4 +1,4 @@
-import sys
+import six
 from pyspark.ml.util import JavaMLReadable, JavaMLWritable, _jvm
 from pyspark.ml.wrapper import JavaTransformer
 from pyspark.ml.param.shared import HasInputCol, HasOutputCol
@@ -26,9 +26,8 @@ class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, Java
             assert handleInvalid in ['error', 'keep'], 'Invalid value for handleInvalid: {}'.format(handleInvalid)
             assert isinstance(labels, dict), 'labels must be a dict or a DataFrame, got: {}'.format(type(labels))
             for (key, value) in labels.items():
-                # py2 compatibility: DataFrame keys are typically unicode
-                key_types = (str) if sys.version_info > (3, 0) else (str, unicode)
-                assert isinstance(key, key_types), 'label keys must be {}, got: {}'.format(key_types, type(key))
+                assert isinstance(key, six.string_types), \
+                    'label keys must be a string type, got: {}'.format(type(key))
                 assert isinstance(value, float), 'label values must be float, got: {}'.format(type(key))
 
         validate_args()
--- a/python/mleap/pyspark/feature/string_map.py
+++ b/python/mleap/pyspark/feature/string_map.py
@@ -1,6 +1,8 @@
+import sys
 from pyspark.ml.util import JavaMLReadable, JavaMLWritable, _jvm
 from pyspark.ml.wrapper import JavaTransformer
 from pyspark.ml.param.shared import HasInputCol, HasOutputCol
+from pyspark.sql import DataFrame
 
 
 class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, JavaMLWritable):
@@ -10,8 +12,27 @@ class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, Java
         labels is a dict {string: double}
         handleInvalid: how to handle missing labels: 'error' (throw an error), or 'keep' (map to the default value)
         """
-        assert handleInvalid in ['error', 'keep'], 'Invalid value for handleInvalid: {}'.format(handleInvalid)
         super(StringMap, self).__init__()
+
+        if isinstance(labels, DataFrame):
+            assert inputCol, 'inputCol is required when labels is a DataFrame'
+            assert outputCol, 'outputCol is required when labels is a DataFrame'
+            labels = {row[0]: float(row[1]) for row in labels.select([inputCol, outputCol]).collect()}
+
+        def validate_args():
+            """
+            validate args early to avoid failing at Py4j with some hard to interpret error message
+            """
+            assert handleInvalid in ['error', 'keep'], 'Invalid value for handleInvalid: {}'.format(handleInvalid)
+            assert isinstance(labels, dict), 'labels must be a dict or a DataFrame, got: {}'.format(type(labels))
+            for (key, value) in labels.items():
+                # py2 compatibility: DataFrame keys are typically unicode
+                key_types = (str) if sys.version_info > (3, 0) else (str, unicode)
+                assert isinstance(key, key_types), 'label keys must be {}, got: {}'.format(key_types, type(key))
+                assert isinstance(value, float), 'label values must be float, got: {}'.format(type(key))
+
+        validate_args()
+
         labels_scala_map = _jvm() \
             .scala \
             .collection \
@@ -26,14 +47,3 @@ class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, Java
         self._java_obj = self._new_java_obj("org.apache.spark.ml.mleap.feature.StringMap", self.uid, string_map_model)
         self.setInputCol(inputCol)
         self.setOutputCol(outputCol)
-
-    @classmethod
-    def from_dataframe(cls, df, key_col, value_col, handleInvalid='error', defaultValue=0.0):
-        """
-        from_dataframe(self, df, key_col, value_col)
-        Creates StringMap from a DataFrame.
-        """
-        dict_from_df = {row[0]: float(row[1]) for row in
-                        df.select([key_col, value_col]).collect()}
-        return cls(dict_from_df, inputCol=key_col, outputCol=value_col, handleInvalid=handleInvalid,
-                   defaultValue=defaultValue)
--- a/python/mleap/pyspark/feature/string_map.py
+++ b/python/mleap/pyspark/feature/string_map.py
@@ -6,7 +6,7 @@ from pyspark.sql import DataFrame
 
 
 class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, JavaMLWritable):
-    def __init__(self, labels, inputCol=None, outputCol=None, handleInvalid='error', defaultValue=0.0):
+    def __init__(self, labels={}, inputCol=None, outputCol=None, handleInvalid='error', defaultValue=0.0):
         """
         :param labels: a dict {string: double}
         :param handleInvalid: how to handle missing labels: 'error' (throw), or 'keep' (map to defaultValue)
--- a/python/tests/pyspark/lib/assertions.py
+++ b/python/tests/pyspark/lib/assertions.py
@@ -11,7 +11,7 @@ def _print_dfs(actual, expected):
 
 def assert_df(actual, expected, sort=True):
     """
-    For more elaborate DFs comparision, e.g. ignoring sorting or less precise.
+    For more elaborate DFs comparison, e.g. ignoring sorting or less precise.
     Default precision for float comparison is 5 digits.
     """
     __tracebackhide__ = True
--- a/python/tests/pyspark/lib/assertions.py
+++ b/python/tests/pyspark/lib/assertions.py
@@ -9,7 +9,7 @@ def _print_dfs(actual, expected):
     expected.show(100, False)
 
 
-def assert_df(actual, expected, sort=True, **kwargs):
+def assert_df(actual, expected, sort=True):
     """
     For more elaborate DFs comparision, e.g. ignoring sorting or less precise.
     Default precision for float comparison is 5 digits.
@@ -21,7 +21,7 @@ def assert_df(actual, expected, sort=True, **kwargs):
         expected = expected.orderBy(actual.columns)
 
     try:
-        assert_frame_equal(actual.toPandas(), expected.toPandas(), **kwargs)
+        assert_frame_equal(actual.toPandas(), expected.toPandas())
     except AssertionError as e:
         _print_dfs(actual, expected)
         # assert_frame_equal doesn't print the column name, only the index
--- a/python/tests/pyspark/lib/assertions.py
+++ b/python/tests/pyspark/lib/assertions.py
@@ -1,0 +1,35 @@
+import re
+from pandas.testing import assert_frame_equal
+
+
+def _print_dfs(actual, expected):
+    print('actual.show(100):')
+    actual.show(100, False)
+    print('expected.show(100):')
+    expected.show(100, False)
+
+
+def assert_df(actual, expected, sort=True, **kwargs):
+    """
+    For more elaborate DFs comparision, e.g. ignoring sorting or less precise.
+    Default precision for float comparison is 5 digits.
+    """
+    __tracebackhide__ = True
+
+    if sort:
+        actual = actual.orderBy(actual.columns)
+        expected = expected.orderBy(actual.columns)
+
+    try:
+        assert_frame_equal(actual.toPandas(), expected.toPandas(), **kwargs)
+    except AssertionError as e:
+        _print_dfs(actual, expected)
+        # assert_frame_equal doesn't print the column name, only the index
+        #   -> get the index from message with regex & resolve column name
+        matcher = re.search(r'iloc\[:, (\d+)\]', e.args[0])
+        if matcher:
+            iloc = matcher.group(1)
+            if iloc is not None:
+                raise AssertionError("failed assert on column '{}': {}".format(actual.columns[int(iloc)], e))
+        # couldn't extract column name. unexpected.
+        raise e
--- a/python/tests/pyspark/lib/spark_session.py
+++ b/python/tests/pyspark/lib/spark_session.py
@@ -30,7 +30,7 @@ def _mleap_classpath():
     """
     Read classpath for mleap-spark-extension with the locally compiled classes.
 
-    Classpath file can be refreshed by running `sbt mleap-spark-extension/compile`.
+    Classpath file can be refreshed by running `sbt mleap-spark-extension/writeRuntimeClasspathToFile`.
     However, that's only needed if making changes to dependencies.
     """
     classpath_file = os.path.join(
--- a/python/tests/pyspark/lib/spark_session.py
+++ b/python/tests/pyspark/lib/spark_session.py
@@ -40,10 +40,10 @@ def _mleap_classpath():
     as all jars are with the same scala version.
 
     Classpath file can be refreshed manually by running `sbt mleap-spark-extension/writeRuntimeClasspathToFile`.
-    However, that's only needed if making changes to dependencies, and any sbt compile runs writeRuntimeClasspathToFile
+    However, that's only needed if making changes to dependencies, and any sbt +compile runs writeRuntimeClasspathToFile
     """
     classpath_file = os.path.join(os.path.dirname(__file__), '..', '..', '..', '..',
                                   'mleap-spark-extension', 'target', 'classpath-runtime_2.11.8.txt')
-    assert os.path.exists(classpath_file), 'classpath-runtime.txt is missing. Run sbt compile first to generate it.'
+    assert os.path.exists(classpath_file), 'classpath-runtime.txt is missing. Run sbt +compile first to generate it.'
     with open(classpath_file, 'r') as f:
         return f.read()
--- a/python/tests/pyspark/lib/spark_session.py
+++ b/python/tests/pyspark/lib/spark_session.py
@@ -1,0 +1,39 @@
+import os
+from pyspark.sql import SparkSession
+
+
+def spark_session():
+    """
+    Minimize parallelism etc. to speed up execution.
+    Test data is small & local, so just avoid all kind of overhead.
+    """
+    builder = SparkSession.builder \
+        .config('spark.sql.shuffle.partitions', 1) \
+        .config('spark.default.parallelism', 1) \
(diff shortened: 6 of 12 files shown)


ancasarb:
@juhoautio I've merged the python/tox changes into master, can you please update the PR, thank you!

juhoautio:
This is also waiting for https://github.com/combust/mleap/pull/569

juhoautio:
@lucagiovagnoli, @lydian (as you have made python contributions in mleap)
\+ @ancasarb

Would you have any comments on the "Open questions" in this PR's description?

lydian:
I honestly prefer pytest myself, so I personally don't mind moving all the tests to pytest.  However, for spark, it is still doable to use single pyspark with unittest. Here's an example of how to: https://github.com/Yelp/mrjob/blob/e826a6b6e5042cd9a6d590d3613268f8428457f4/tests/sandbox.py#L188  

I am not sure about the second question. Although we are using tox, the mleap is not in the requirements.txt or dependency. Which means it will just search for the local directory and found the mleap folder and use that mleap folder to run test.  Therefore you shouldn't have issue on debugging with your own code using tox.  Let me know what the issue you're facing, and how I can better helping :) 

I'll leave the last question to @lucagiovagnoli  since he is more experienced on connecting scala and python.

python/tox.ini
```diff
@@ -0,0 +1,22 @@
+[tox]
+envlist = py27,py36,py37
+skipdist = true
+
+[testenv]
+deps =
+        -rrequirements.txt
```
lucagiovagnoli: I just added -rrequirements.txt into the requirements-dev.txt file. So we reference only the dev file in tox.ini

mleap-executor-testkit/src/main/scala/ml/combust/mleap/executor/testkit/TransformServiceSpec.scala
```diff
@@ -361,7 +361,8 @@ trait TransformServiceSpec extends FunSpecLike
       }
 
       describe("throttling") {
-        it("throttles elements through the flow") {
+        // TODO this test fails randomly on Travis, so temporarily replaced it() with ignore()
```
lucagiovagnoli: Do you have a link to an example of a build failure ? I can take a quick look

mleap-spark/build.sbt
```diff
@@ -2,3 +2,12 @@ import ml.combust.mleap.{Dependencies, Common}
 
 Common.defaultMleapSettings
 Dependencies.spark
+
+val printDependencyClasspath: TaskKey[Unit] = taskKey[Unit]("Prints locations of the dependencies")
+
+printDependencyClasspath := {
+  val cp = (dependencyClasspath in Compile).value
+  // TODO write to file so that this can be read by pyspark tests? or how should this work, maybe pyspark tests should
```
lucagiovagnoli: I'm not sure what you are trying to do here so I am not sure how to answer question 3) in this PR

python/tox.ini
juhoautio: Thanks. I will update this eventually once https://github.com/combust/mleap/pull/569 has been merged.

python/mleap/pyspark/feature/string_map.py
```diff
@@ -0,0 +1,46 @@
+from pyspark.ml.util import JavaMLReadable, JavaMLWritable, _jvm
+from pyspark.ml.wrapper import JavaTransformer
+from pyspark.ml.param.shared import HasInputCol, HasOutputCol
+from pyspark.sql import DataFrame
+
+
+class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, JavaMLWritable):
```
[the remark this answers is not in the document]
juhoautio: Uh, yeah this PR is now maybe too outdated :) I'll be rebasing this..

mleap-executor-testkit/src/main/scala/ml/combust/mleap/executor/testkit/TransformServiceSpec.scala
juhoautio: @lucagiovagnoli 

https://github.com/combust/mleap/pull/568#issuecomment-532371321
https://github.com/combust/mleap/pull/568#discussion_r327019642

This commit was actually merged as part of https://github.com/combust/mleap/pull/568 so will disappear from this diff after rebasing.

Should continue discussion about the flake at https://github.com/combust/mleap/pull/568

mleap-spark/build.sbt
juhoautio: @lucagiovagnoli thanks, what I actually want to achieve is importing the contents of mleap jar (and dependencies) in the SparkSession of pyspark tests, but in a quick & easy-to-develop way. When using a published mleap from maven, it's simple:

```python
SparkSession.builder \
    .config("spark.jars.packages", "ml.combust.mleap:mleap-spark-extension_2.11:0.14.0")
```

However, in development mode I want to use the local version of mleap code.

I found that it's possible to make it work by listing the paths to all classes/ folders of locally built mleap & dependency jars. However, this is kinda manual. I was wondering if sbt would provide some easier way to export the required runtime classpath (when sbt is for example running tests locally) for the `mleap-spark-extension` module.

Another way to do this could be to install mleap to local maven repo with with some snapshot version and use it with `"spark.jars.packages"` (but obviously that's a bit slow for testing new changes, and not too clean, because it's writing to your local maven repo).

juhoautio: Ok, here's a clean way to get the classpath on command line:

`sbt "export mleap-spark-extension/runtime:fullClasspath"`

However it prints some additional log lines first.. Maybe I will still look into a programmatic way.

juhoautio:
@ancasarb what do you say about using pytest in mleap? At least initially it could be used only for the pyspark tests.

@lydian thanks for your comment:

> it is still doable to use single pyspark with unittest. Here's an example of how to: https://github.com/Yelp/mrjob/blob/e826a6b6e5042cd9a6d590d3613268f8428457f4/tests/sandbox.py#L188

If I understood correctly, it requires all this code:

https://github.com/Yelp/mrjob/blob/e826a6b6e5042cd9a6d590d3613268f8428457f4/tests/sandbox.py#L188-L228

And also all test cases need to inherit the `SingleSparkContextTestCase` class?

Clearly not ideal.

In pytest spark_session could be defined as fixture that's scoped for the entire testing session with a single line, and that's it!

juhoautio:
@ancasarb now ready for review/merge

ancasarb:
I will take a look this evening

python/mleap/pyspark/feature/string_map.py
```diff
@@ -26,14 +47,3 @@ def __init__(self, labels, inputCol=None, outputCol=None, handleInvalid='error',
         self._java_obj = self._new_java_obj("org.apache.spark.ml.mleap.feature.StringMap", self.uid, string_map_model)
         self.setInputCol(inputCol)
         self.setOutputCol(outputCol)
-
-    @classmethod
-    def from_dataframe(cls, df, key_col, value_col, handleInvalid='error', defaultValue=0.0):
```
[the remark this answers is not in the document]
juhoautio: This is cleaner, because there's no need to echo the same kwargs as required by the constructor.

I also wrote this in PR desc about it:

> Remove separate from_dataframe method. Allow passing a DataFrame as an alternative to a dict in the same constructor.

python/tests/pyspark/lib/assertions.py
```diff
@@ -0,0 +1,35 @@
...
+def _print_dfs(actual, expected):
+    print('actual.show(100):')
+    actual.show(100, False)
+    print('expected.show(100):')
+    expected.show(100, False)
+
+
+def assert_df(actual, expected, sort=True, **kwargs):
+    """
+    For more elaborate DFs comparision, e.g. ignoring sorting or less precise.
```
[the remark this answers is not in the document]
juhoautio: assert_frame_equal seems to do the job: it compares floats with restricted precision instead of full. Or am I missing how allclose would handle it better / easier?

.travis.yml
```diff
@@ -1,41 +1,43 @@
...
-jdk:
-  - oraclejdk8
-cache:
-  directories:
-    - $HOME/.ivy2/cache
-    - $HOME/.sbt/boot/
-before_cache:
-  - find $HOME/.ivy2 -name "ivydata-*.properties" -delete
-  - find $HOME/.sbt -name "*.lock" -delete
+# Use container-based infrastructure
```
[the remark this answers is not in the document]
juhoautio: Seems that file in master has CRLF's. My file has LF's. Viewing the diff with whitespace ignored shows "actual" changes lines.

python/tests/pyspark/lib/spark_session.py
```diff
@@ -0,0 +1,39 @@
+import os
+from pyspark.sql import SparkSession
+
+
+def spark_session():
```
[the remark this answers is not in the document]
juhoautio: Yes, as I wrote above: https://github.com/combust/mleap/pull/571#issuecomment-535701643

python/mleap/pyspark/feature/string_map.py
```diff
@@ -1,17 +1,38 @@
...
+
+        def validate_args():
+            """
+            validate args early to avoid failing at Py4j with some hard to interpret error message
+            """
+            assert handleInvalid in ['error', 'keep'], 'Invalid value for handleInvalid: {}'.format(handleInvalid)
+            assert isinstance(labels, dict), 'labels must be a dict or a DataFrame, got: {}'.format(type(labels))
+            for (key, value) in labels.items():
+                # py2 compatibility: DataFrame keys are typically unicode
+                key_types = (str) if sys.version_info > (3, 0) else (str, unicode)
```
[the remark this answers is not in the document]
juhoautio: So, `import six` & check for `isinstance(value, six.string_types)`?

mleap-executor-testkit/src/main/scala/ml/combust/mleap/executor/testkit/TransformServiceSpec.scala
```diff
@@ -232,7 +232,8 @@ trait TransformServiceSpec extends FunSpecLike
...
```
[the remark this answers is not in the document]
juhoautio: https://travis-ci.org/combust/mleap/builds/590427424#L4478-L4493

```
[info]   transform row stream
4479[info]   - transforms rows in a stream
4480[info]     throttling
4481[info]     - throttles elements through the flow *** FAILED ***
4482[info]       java.lang.AssertionError: assertion failed: expected OnNext(2), found OnNext(3)
```

.travis.yml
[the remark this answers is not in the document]
lucagiovagnoli: oh ok, makes sense

python/mleap/pyspark/feature/string_map.py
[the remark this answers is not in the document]
juhoautio: This is great. Perfectly cleaned up by using six.

juhoautio:
Rebased & force-pushed.

mleap-executor-testkit/src/main/scala/ml/combust/mleap/executor/testkit/TransformServiceSpec.scala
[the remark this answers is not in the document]
lucagiovagnoli: Is there a github issue tackling this flake test ?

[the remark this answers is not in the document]
ancasarb: I'll create one / raise a PR to fix this, sorry for the build issues created by this flaky one.

python/mleap/pyspark/feature/string_map.py
```diff
@@ -1,17 +1,37 @@
...
+        if isinstance(labels, DataFrame):
+            assert inputCol, 'inputCol is required when labels is a DataFrame'
+            assert outputCol, 'outputCol is required when labels is a DataFrame'
+            labels = {row[0]: float(row[1]) for row in labels.select([inputCol, outputCol]).collect()}
+
+        def validate_args():
+            """
+            validate args early to avoid failing at Py4j with some hard to interpret error message
+            """
+            assert handleInvalid in ['error', 'keep'], 'Invalid value for handleInvalid: {}'.format(handleInvalid)
```
lucagiovagnoli: you could make handleInvalid be an Enum, rather than manually checking here

python/mleap/pyspark/feature/string_map.py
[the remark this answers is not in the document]
lucagiovagnoli: Repeating args is not a big problem. I'd argue that the @classmethod is a more pythonic way to handle multiple different constructor. 
This is an example of the multiple constructor pattern via @classmethod: https://stackoverflow.com/a/682545

A problem is that `labels` becomes a variable holding objects of different types (generally not even allowed in other languages) that is python strength and curse at the same time.

Out unique constructor includes now an if statement  (`if isinstance(labels, DataFrame):`) breaking the [Principle of Single Responsibility](https://en.wikipedia.org/wiki/Single_responsibility_principle) for our __init__ function. We also get a hidden behaviour (a special handling of Dataframes) in the __init__ that goes against transparency.

Python doesn't have typed values so sometimes we must be explicit with naming. Take this example:
```python
some_data = DictOrDataframe()

# reader doesn know if some_data is a dict or a dataframe
StringMap(some_data)

# reader knows some_data is a dataframe
StringMap.from_dataframe(some_data)
```

mleap-executor-testkit/src/main/scala/ml/combust/mleap/executor/testkit/TransformServiceSpec.scala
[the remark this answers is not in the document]
lucagiovagnoli: Thanks Anca! If we ignore this test, it'd be nice to get the issue number in this comment

python/tests/pyspark/lib/spark_session.py
```diff
@@ -0,0 +1,39 @@
...
+
+def _mleap_classpath():
+    """
+    Read classpath for mleap-spark-extension with the locally compiled classes.
+
+    Classpath file can be refreshed by running `sbt mleap-spark-extension/writeRuntimeClasspathToFile`.
+    However, that's only needed if making changes to dependencies.
+    """
+    classpath_file = os.path.join(
+        os.path.dirname(__file__), '..', '..', '..', '..', 'mleap-spark-extension', 'target', 'classpath-runtime.txt')
```
lucagiovagnoli: Building a path this way looks brittle, what happens if someone runs the tests from a different folder? 
I'd aim for an absolute classpath. Also It'd be great if you could pass this classpath as an environment variable rather than a file somewhere on disk

mleap-spark-extension/build.sbt
```diff
@@ -2,3 +2,18 @@ import ml.combust.mleap.{Dependencies, Common}
 
 Common.defaultMleapSettings
 Dependencies.sparkExtension
+
+val writeRuntimeClasspathToFile = taskKey[Unit]("Writes runtime classpath to a file")
+
+writeRuntimeClasspathToFile := {
```
lucagiovagnoli: Sorry for my late answer to question 3! I had to look into this.

If we need to pass the classpath to python tests, have you considered to pass an environment variable ? Execution would look like:
```
MLEAP_CLASSPATH='/some/folder/with/classpath/' tox -e py36
```

In order to generate the classpath, I think there are some useful sbt commands, this [Stackoverflow answer](https://stackoverflow.com/a/21738753) shows some useful ones.

python/mleap/pyspark/feature/string_map.py
juhoautio: Sorry, but what Enum? Something that works for all envs: py27,py36,py37 (https://github.com/combust/mleap/blob/master/python/tox.ini#L2)?

python/tests/pyspark/lib/spark_session.py
juhoautio: > what happens if someone runs the tests from a different folder?

It always works. Because `__file__` is used to refer to `classpath-runtime.txt` with a _relative_ path.

juhoautio: > an environment variable rather than a file somewhere on disk

I think caching in a file is a must. For example, when I'm developing a pyspark test I don't want the additional overhead of invoking sbt every time that I run the test again.

For me it takes ~10s to run either of:
- `sbt mleap-spark-extension/writeRuntimeClasspathToFile`
- `sbt mleap-spark-extension/runtime:fullClasspath`

mleap-spark-extension/build.sbt
juhoautio: My goal was this: export classpath to a file as part of any regular sbt compile that affects mleap-spark-extension.

Having sbt task that does it is the first step – check.

Problem 1: I failed to find a way to make any compile (regular compile vs. test:compile) include this task. For example `sbt "inspect tree mleap-spark-extension/test:compile"` showed that it doesn't depend on `writeRuntimeClasspathToFile`.

Travis is instructed to run `sbt test`. I don't really know how this works with sbt, but `test:compile` also compiles the runtime sources, not only test sources. However doesn't execute the regular `compile:compile` at all and so making that depend on `writeRuntimeClasspathToFile` won't make `sbt test` depend on it.

----

Otherwise this produces the same output as my custom `writeRuntimeClasspathToFile` task:

`sbt --error "export mleap-spark-extension/runtime:fullClasspath" > fullClasspath.txt`

(`--error` is needed to disable info logging printed into the the file as well)

Problem 2: I couldn't find how to run the same "command" as scala from sbt file, but I guess it should be doable. That would allow simplifying the "impl" of the writeRuntimeClasspathToFile task.

mleap-executor-testkit/src/main/scala/ml/combust/mleap/executor/testkit/TransformServiceSpec.scala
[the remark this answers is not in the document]
juhoautio: Here's a github issue:
https://github.com/combust/mleap/issues/573

mleap-spark-extension/build.sbt
juhoautio: > have you considered to pass an environment variable

I didn't. If we go there, I would prefer it to be done in a IDE-compatible way, not something that is set in the `Makefile` or some travis conf file and only works when executing tests through them.

If switching to pytest, I know a clean way at least:
- set the env var in `conftest.py` with `os.environ` in `def pytest_generate_tests(metafunc)`
- that way env var is available always when running `pytest` (IDE or command line) without any manual setup

In general I would like this PR's scope to remain small enough, and then improve later with switching to pytest, if anyone is up for that.

python/mleap/pyspark/feature/string_map.py
lucagiovagnoli: I was thinking this one https://cpython-test-docs.readthedocs.io/en/latest/library/enum.html but it's more of a nice to have. We can also think about this again after 2020 when python 2.7 support is completely dropped (maybe eventually dropping it here too..).

mleap-executor-testkit/src/main/scala/ml/combust/mleap/executor/testkit/TransformServiceSpec.scala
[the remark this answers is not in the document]
lucagiovagnoli: Thanks ! Nice job. I 'approved' the PR #574

juhoautio:
@lucagiovagnoli would you have time to continue with this?

lucagiovagnoli:
I'll have another look later today, there's a few things I don't feel fully sure about

python/tests/pyspark/lib/spark_session.py
lucagiovagnoli: You don't have to invoke sbt every time you run the tests, you could run it once and copy paste the output to an env. variable or your personal file on disk as it fits best your development flow. This code here makes a very strong move towards a way to develop and run tests that goes through a file-cache. Not everyone might use a similar environment, IDE or development-process, that's why I am suggesting to leave this less opinionated and flexible (for example I don't use an IDE). I'd rather let devs run sbt once, then they can paste it wherever it fits best for them

python/tests/pyspark/lib/assertions.py
```diff
@@ -0,0 +1,35 @@
...
+        actual = actual.orderBy(actual.columns)
+        expected = expected.orderBy(actual.columns)
+
+    try:
+        assert_frame_equal(actual.toPandas(), expected.toPandas(), **kwargs)
+    except AssertionError as e:
+        _print_dfs(actual, expected)
+        # assert_frame_equal doesn't print the column name, only the index
+        #   -> get the index from message with regex & resolve column name
+        matcher = re.search(r'iloc\[:, (\d+)\]', e.args[0])
```
lucagiovagnoli: I'd avoid using regexes in code. They are hard to read and maintain: https://stackoverflow.com/questions/764247/why-are-regular-expressions-so-controversial

python/tests/pyspark/lib/assertions.py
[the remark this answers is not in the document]
lucagiovagnoli: So why the comment here says "less precise" if assert_frame_equal does that already? I thought this testing function tried to introduce less precise assertions. Is this testing meant to just print out columns more clearly? I've never found this to be an issue for assert_frame_equal. Do you have an  example of this new output vs. the output of assert_frame_equal ?

python/tests/pyspark/lib/assertions.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def _print_dfs(actual, expected):
+    print('actual.show(100):')
+    actual.show(100, False)
+    print('expected.show(100):')
+    expected.show(100, False)
+
+
+def assert_df(actual, expected, sort=True, **kwargs):
```
lucagiovagnoli: kwargs are generally dangerous and hard to document. Which kwargs does assert_df accepts ? By reading the function code I think they are the same kwargs as assert_frame_equal but this is not immediate from the function signature

python/mleap/pyspark/feature/string_map.py
juhoautio: So, resolve this conversion maybe?

python/tests/pyspark/lib/spark_session.py
juhoautio: I'm not fully sure what kind of flow this would be:

> copy paste the output to an env. variable or your personal file on disk

My aim was to make the tests work out of the without manual tweaking or configuring. Do you have some idea how that could be achieved otherwise?

Pyspark tests will any way depend on some generated files (classes), so is it any worse to write a file that has the required classpath?

mleap-spark-extension/build.sbt
lucagiovagnoli: Aren't env. variables IDE-compatible? they can generally be set in the configuration section and then used in the testing environment.

How's the file cleaned up? What if a testing machine has write-permissions disabled or if I run sbt from a different folder?  I'd rather have a sbt task printing `dependencyClasspath in Runtime` to screen and then do things with it.

Also happy to use pytest if possible.

python/tests/pyspark/lib/assertions.py
juhoautio: The reason for having this is, as the comment says:

> assert_frame_equal doesn't print the column name, only the index

Should this use some other way for capturing the column name from the string than regex?

python/tests/pyspark/lib/assertions.py
juhoautio: Yes, those are simply passed on to assert_frame_equal. That should be added to docstring here.

I think passing kwargs is more future-proof than copying all of these and trying to keep them up-to-date:

```
def assert_frame_equal(left, right, check_dtype=True,
                       check_index_type='equiv',
                       check_column_type='equiv',
                       check_frame_type=True,
                       check_less_precise=False,
                       check_names=True,
                       by_blocks=False,
                       check_exact=False,
                       check_datetimelike_compat=False,
                       check_categorical=True,
                       check_like=False,
                       obj='DataFrame'):
```

On the other hand as long as nothing needs to be passed kwargs can be just removed.

python/tests/pyspark/lib/assertions.py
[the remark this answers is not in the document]
juhoautio: I can remove "less precise". There's no context for that left here, but used to be vs. direct comparison like this, for example:

```python
        assert sorted(list(actual.na.fill('null').toLocalIterator())) == \
               sorted(list(expected.na.fill('null').toLocalIterator()))
```

^ This would be too precise and easily fail on floating point differences.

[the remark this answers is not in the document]
juhoautio: > Is this testing meant to just print out columns more clearly?

Not only that. This also sorts the df's (default sort=True) before comparing with assert_frame_equal.

> Do you have an example of this new output vs. the output of assert_frame_equal ?

I modified the `test_map` here to test.

Output with assert_df:

```
actual.show(100):
+-------+---------+---------+
|key_col|extra_col|value_col|
+-------+---------+---------+
|a      |b        |2.0      |
+-------+---------+---------+

expected.show(100):
+-------+---------+---------+
|key_col|extra_col|value_col|
+-------+---------+---------+
|a      |b        |1.0      |
+-------+---------+---------+



Ran 1 test in 10.802s

FAILED (failures=1)

Failure
Traceback (most recent call last):
  File "/Users/JuhoAutio/ideaprojects/mleap/python/tests/pyspark/lib/assertions.py", line 24, in assert_df
    assert_frame_equal(actual.toPandas(), expected.toPandas(), **kwargs)
  File "/Users/JuhoAutio/ideaprojects/mleap/python/venv/lib/python3.7/site-packages/pandas/util/testing.py", line 1502, in assert_frame_equal
    obj='DataFrame.iloc[:, {idx}]'.format(idx=i))
  File "/Users/JuhoAutio/ideaprojects/mleap/python/venv/lib/python3.7/site-packages/pandas/util/testing.py", line 1336, in assert_series_equal
    obj='{obj}'.format(obj=obj))
  File "pandas/_libs/testing.pyx", line 66, in pandas._libs.testing.assert_almost_equal
  File "pandas/_libs/testing.pyx", line 180, in pandas._libs.testing.assert_almost_equal
  File "/Users/JuhoAutio/ideaprojects/mleap/python/venv/lib/python3.7/site-packages/pandas/util/testing.py", line 1094, in raise_assert_detail
    raise AssertionError(msg)
AssertionError: DataFrame.iloc[:, 2] are different

DataFrame.iloc[:, 2] values are different (100.0 %)
[left]:  [2.0]
[right]: [1.0]

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/JuhoAutio/.pyenv/versions/3.7.2/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/Users/JuhoAutio/.pyenv/versions/3.7.2/lib/python3.7/unittest/case.py", line 615, in run
    testMethod()
  File "/Users/JuhoAutio/ideaprojects/mleap/python/tests/pyspark/string_map_test.py", line 29, in test_map
    assert_df(expected, result)
  File "/Users/JuhoAutio/ideaprojects/mleap/python/tests/pyspark/lib/assertions.py", line 33, in assert_df
    raise AssertionError("failed assert on column '{}': {}".format(actual.columns[int(iloc)], e))
AssertionError: failed assert on column 'value_col': DataFrame.iloc[:, 2] are different

DataFrame.iloc[:, 2] values are different (100.0 %)
[left]:  [2.0]
[right]: [1.0]
```

Output with `assert_frame_equal(expected.toPandas(), result.toPandas())`:

```
Traceback (most recent call last):
  File "/Users/JuhoAutio/.pyenv/versions/3.7.2/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/Users/JuhoAutio/.pyenv/versions/3.7.2/lib/python3.7/unittest/case.py", line 615, in run
    testMethod()
  File "/Users/JuhoAutio/ideaprojects/mleap/python/tests/pyspark/string_map_test.py", line 32, in test_map
    assert_frame_equal(expected.toPandas(), result.toPandas())
  File "/Users/JuhoAutio/ideaprojects/mleap/python/venv/lib/python3.7/site-packages/pandas/util/testing.py", line 1502, in assert_frame_equal
    obj='DataFrame.iloc[:, {idx}]'.format(idx=i))
  File "/Users/JuhoAutio/ideaprojects/mleap/python/venv/lib/python3.7/site-packages/pandas/util/testing.py", line 1336, in assert_series_equal
    obj='{obj}'.format(obj=obj))
  File "pandas/_libs/testing.pyx", line 66, in pandas._libs.testing.assert_almost_equal
  File "pandas/_libs/testing.pyx", line 180, in pandas._libs.testing.assert_almost_equal
  File "/Users/JuhoAutio/ideaprojects/mleap/python/venv/lib/python3.7/site-packages/pandas/util/testing.py", line 1094, in raise_assert_detail
    raise AssertionError(msg)
AssertionError: DataFrame.iloc[:, 2] are different

DataFrame.iloc[:, 2] values are different (100.0 %)
[left]:  [2.0]
[right]: [1.0]
```

Now, the actual difference is easy enough to spot when the df's are small (not too many columns or rows).

However, I must say that even this is quite exotic. What if I would have two columns with some numeric values that I want to compare? It wouldn't be _that_ easy to say based on the index number in `DataFrame.iloc[:, 2]` where the problem lies. It doesn't print the column name `value_col` at all, and it doesn't print the full dataframes either.

Maybe there is some better alternative out there so that there would be no need for any of this custom assertion code(?).

mleap-spark-extension/build.sbt
juhoautio: > Aren't env. variables IDE-compatible? they can generally be set in the configuration section and then used in the testing environment.

Yes, it can be done, but is a manual step. I'd really like this to work out of the box.

> How's the file cleaned up?

It doesn't need to be cleaned up, but if I'm not mistaken it would be cleaned up as part of `sbt clean`. Once again, there's an analogy to the .class files from build output.

> What if a testing machine has write-permissions disabled

I don't see how it could then compile the classes.

> I run sbt from a different folder

I don't think sbt can be run from any other folder than root. Or can it? If yes, it should be easy to get the module's build output folder in build.sbt, so that the path for classpath-runtime.txt can be relative to that.

> I'd rather have a sbt task printing dependencyClasspath in Runtime to screen and then do things with it.

I really don't understand why that would be better. Sorry :/ To me it just seems like redundant manual steps to run the tests.

> Also happy to use pytest if possible.

Definitely, but not part of this PR, please. AFAIK pytest doesn't make it easier to have the classpath available for the tests. It's more about invoking sbt.

juhoautio:
Thanks for looking into this @lucagiovagnoli !

juhoautio:
@lucagiovagnoli @ancasarb when you have time, some open questions here ^

mleap-spark-extension/build.sbt
juhoautio: > My goal was this: export classpath to a file as part of any regular sbt compile that affects mleap-spark-extension.

> Problem 1: I failed to find a way to make any compile (regular compile vs. test:compile) include this task

I found a way by using sbt's triggeredBy. Committed it here: https://github.com/combust/mleap/pull/571/commits/4a0ce0738ab19ca45ca459d6686d689cd54321f8

juhoautio:
Whee, now I'm finally happy with this!

juhoautio:
Added commit:

Fix deserialization of StringMap in pyspark wrapper
- Add a unit test for pyspark de/serialization roundtrip of StringMap
- Map the packages in the way that works together with pyspark's wrapper deserialization
- Make the `labels` arg optional in StringMap constructor so that pyspark's py_type() call (no args) in deserialization works

The fix itself could be included separately as an individual PR, but it's of course nice to accompany it with the unit test like here (which is only possible if the pyspark test tooling is in place..).

The fixing imports part in `python/mleap/pyspark/__init__.py` is quite a hack. Well, at least it makes the test work! I wonder what's the need for all these mapping with the artificial `pyspark.ml` package prefix in the first place?

python/mleap/pyspark/__init__.py
```diff
@@ -1,5 +1,13 @@
+import pyspark.ml
 import mleap.pyspark
+import mleap.pyspark.feature
+from mleap.pyspark.feature.string_map import StringMap
 import sys
 
 sys.modules['pyspark.ml.mleap'] = mleap
 sys.modules['pyspark.ml.mleap.pyspark'] = sys.modules['mleap.pyspark']
+sys.modules['pyspark.ml.mleap.feature'] = sys.modules['mleap.pyspark.feature']
```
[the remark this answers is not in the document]
juhoautio: Yes, tests.pyspark.string_map_test.StringMapTest.test_serialize_to_bundle fails if this is removed:

```
Error
Traceback (most recent call last):
  File "/Users/juhoautio/.pyenv/versions/2.7.17/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/Users/juhoautio/ideaprojects/mleap/python/tests/pyspark/string_map_test.py", line 63, in test_serialize_to_bundle
    deserialized_pipeline = _deserialize_from_file(pipeline_file)
  File "/Users/juhoautio/ideaprojects/mleap/python/tests/pyspark/string_map_test.py", line 103, in _deserialize_from_file
    return SimpleSparkSerializer().deserializeFromBundle(_to_file_path(path))
  File "/Users/juhoautio/ideaprojects/mleap/python/mleap/pyspark/spark_support.py", line 45, in deserializeFromBundle
    return JavaTransformer._from_java( self._java_obj.deserializeFromBundle(path))
  File "/Users/juhoautio/ideaprojects/mleap/python/.tox/py27/lib/python2.7/site-packages/pyspark/ml/wrapper.py", line 240, in _from_java
    py_stage = py_type._from_java(java_stage)
  File "/Users/juhoautio/ideaprojects/mleap/python/.tox/py27/lib/python2.7/site-packages/pyspark/ml/pipeline.py", line 299, in _from_java
    py_stages = [JavaParams._from_java(s) for s in java_stage.stages()]
  File "/Users/juhoautio/ideaprojects/mleap/python/.tox/py27/lib/python2.7/site-packages/pyspark/ml/wrapper.py", line 227, in _from_java
    py_type = __get_class(stage_name)
  File "/Users/juhoautio/ideaprojects/mleap/python/.tox/py27/lib/python2.7/site-packages/pyspark/ml/wrapper.py", line 221, in __get_class
    m = __import__(module)
ImportError: No module named feature
```

python/mleap/pyspark/feature/string_map.py
```diff
@@ -1,17 +1,36 @@
...
+            assert isinstance(labels, dict), 'labels must be a dict, got: {}'.format(type(labels))
+            for (key, value) in labels.items():
+                assert isinstance(key, six.string_types), \
+                    'label keys must be a string type, got: {}'.format(type(key))
+                assert isinstance(value, float), 'label values must be float, got: {}'.format(type(key))
```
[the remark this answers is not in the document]
juhoautio: What kind of double? Python only knows `float` as a primitive type?

python/mleap/pyspark/__init__.py
```diff
@@ -1,5 +1,13 @@
...
+
+sys.modules['pyspark.ml'].mleap = mleap
+sys.modules['pyspark.ml'].mleap.feature = sys.modules['mleap.pyspark.feature']
+sys.modules['pyspark.ml'].mleap.feature.StringMap = StringMap
```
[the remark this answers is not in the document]
juhoautio: You probably know better than me why these mappings are needed. I "learned" about this hack from the history of this repo :)

Any way, tests.pyspark.string_map_test.StringMapTest.test_serialize_to_bundle fails if this line is removed:

```
Error
Traceback (most recent call last):
  File "/Users/juhoautio/.pyenv/versions/2.7.17/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/Users/juhoautio/ideaprojects/mleap/python/tests/pyspark/string_map_test.py", line 63, in test_serialize_to_bundle
    deserialized_pipeline = _deserialize_from_file(pipeline_file)
  File "/Users/juhoautio/ideaprojects/mleap/python/tests/pyspark/string_map_test.py", line 103, in _deserialize_from_file
    return SimpleSparkSerializer().deserializeFromBundle(_to_file_path(path))
  File "/Users/juhoautio/ideaprojects/mleap/python/mleap/pyspark/spark_support.py", line 45, in deserializeFromBundle
    return JavaTransformer._from_java( self._java_obj.deserializeFromBundle(path))
  File "/Users/juhoautio/ideaprojects/mleap/python/.tox/py27/lib/python2.7/site-packages/pyspark/ml/wrapper.py", line 240, in _from_java
    py_stage = py_type._from_java(java_stage)
  File "/Users/juhoautio/ideaprojects/mleap/python/.tox/py27/lib/python2.7/site-packages/pyspark/ml/pipeline.py", line 299, in _from_java
    py_stages = [JavaParams._from_java(s) for s in java_stage.stages()]
  File "/Users/juhoautio/ideaprojects/mleap/python/.tox/py27/lib/python2.7/site-packages/pyspark/ml/wrapper.py", line 227, in _from_java
    py_type = __get_class(stage_name)
  File "/Users/juhoautio/ideaprojects/mleap/python/.tox/py27/lib/python2.7/site-packages/pyspark/ml/wrapper.py", line 223, in __get_class
    m = getattr(m, comp)
AttributeError: 'module' object has no attribute 'StringMap'
```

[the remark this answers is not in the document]
juhoautio: In general I mentioned in the commit https://github.com/combust/mleap/pull/571/commits/82d934a59372f6f484589835695873ef7ce3e065:

> - Map the packages in the way that works together with pyspark's wrapper deserialization

For some reason pyspark (I think) changes the module names line this. But I have no idea why.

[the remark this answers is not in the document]
lucagiovagnoli: I believe pyspark does that here: https://github.com/apache/spark/blob/master/python/pyspark/ml/wrapper.py#L243

python/tests/pyspark/string_map_test.py
```diff
@@ -0,0 +1,103 @@
...
+        pipeline = Pipeline(stages=[string_map]).fit(self.input)
+        pipeline_file = os.path.join(os.path.dirname(__file__), '..', '..',
+                                     'target', 'test_serialize_to_bundle-pipeline.zip')
+        _serialize_to_file(pipeline_file, self.input, pipeline)
+        deserialized_pipeline = _deserialize_from_file(pipeline_file)
+        result = deserialized_pipeline.transform(self.input)
+        expected = self.spark.createDataFrame([['a', 'b', 1.0]], OUTPUT_SCHEMA)
+        assert_df(expected, result)
+
+    @staticmethod
```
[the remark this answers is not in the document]
juhoautio: because `self` is not needed in this test, as there is no assertion. lint tools were complaining.

python/tests/pyspark/string_map_test.py
```diff
@@ -0,0 +1,103 @@
...
+            StringMap({'z': 1.0}, 'key_col', 'value_col').transform(self.input).collect()
+        self.assertIn('java.util.NoSuchElementException: Missing label: a', str(error.exception))
+
+    def test_map_from_dataframe(self):
+        labels_df = self.spark.createDataFrame([['a', 1.0]], 'key_col: string, value_col: double')
+        result = StringMap.from_dataframe(labels_df, 'key_col', 'value_col').transform(self.input)
+        expected = self.spark.createDataFrame([['a', 'b', 1.0]], OUTPUT_SCHEMA)
+        assert_df(expected, result)
+
+    @unittest.skip("Works locally but fails on Travis. Don't know how to reproduce")
```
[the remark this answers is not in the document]
juhoautio: There's a PR: https://github.com/combust/mleap/pull/624

python/tests/pyspark/string_map_test.py
```diff
@@ -0,0 +1,103 @@
...
+    def test_serialize_to_bundle(self):
+        string_map = StringMap({'a': 1.0}, 'key_col', 'value_col')
+        pipeline = Pipeline(stages=[string_map]).fit(self.input)
+        pipeline_file = os.path.join(os.path.dirname(__file__), '..', '..',
```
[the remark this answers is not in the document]
juhoautio: @lucagiovagnoli I gave it a stab but with python 2.7 I got a bit lost how to use it for this case. 

`with tempfile.TemporaryDirectory() as temp_dir` would've been perfect, but that's not in Python 2.7.

python/mleap/pyspark/__init__.py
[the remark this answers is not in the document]
talalryz: instead of using sys.modules to monkeypatch pyspark.ml.mleap, can you use the approach here:
`https://github.com/combust/mleap/blob/master/python/mleap/pyspark/spark_support.py#L32`
^This approach is used widely in mleap, and a lot less unwieldy than `sys.modules[..]=...` imo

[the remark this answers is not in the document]
juhoautio: @talalryz thanks, but isn't this different than the sklearn case because there sklearn is patched with setattr, while here the target for patching is pyspark?

Maybe there is some nicer way around this, but the problem here is that pyspark rewrites /maps the module name, and I suppose that doesn't happen in sklearn:
https://github.com/apache/spark/blob/master/python/pyspark/ml/wrapper.py#L243

I was trying to use setattr instead but this was the "best" I could get to work so far:

```python
import pyspark.ml
import mleap.pyspark
import mleap.pyspark.feature
from mleap.pyspark.feature.string_map import StringMap
import sys

sys.modules['pyspark.ml.mleap'] = mleap
sys.modules['pyspark.ml.mleap.pyspark'] = pyspark
sys.modules['pyspark.ml.mleap.feature'] = feature

setattr(pyspark.ml, 'mleap', mleap)
setattr(pyspark.ml.mleap, 'feature', feature)
setattr(pyspark.ml.mleap.feature, 'StringMap', StringMap)
```

Same number of lines but now mixing both sys.modules & setattr – doesn't seem worth changing to that at least.

python/mleap/pyspark/feature/string_map.py
```diff
@@ -1,17 +1,36 @@
+import six
 from pyspark.ml.util import JavaMLReadable, JavaMLWritable, _jvm
 from pyspark.ml.wrapper import JavaTransformer
 from pyspark.ml.param.shared import HasInputCol, HasOutputCol
+from pyspark.sql import DataFrame
 
 
 class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, JavaMLWritable):
-    def __init__(self, labels, inputCol=None, outputCol=None, handleInvalid='error', defaultValue=0.0):
+    def __init__(self, labels={}, inputCol=None, outputCol=None, handleInvalid='error', defaultValue=0.0):
```
lucagiovagnoli, line 9: I'm not sure that all of these parameters are preserved after serialization, have you tested if `handleInvalid` and `defaultValue` are the same after deserializing ?

I don't see how the python `StringMap` can be initialized correctly after deserialization because those 2 fields are not serialized in the first place. They are not part of the StringMap JavaTransformer here: https://github.com/combust/mleap/blob/d276c1c800277a1edcf30dda6e8559c7413b21f8/mleap-spark-extension/src/main/scala/org/apache/spark/ml/mleap/feature/StringMap.scala so I think the just get lost and the deserialized object will always have the defaults `error` and 0.0

(I also see the deserialization test is inactive, see https://github.com/combust/mleap/pull/624)

juhoautio: @lucagiovagnoli thanks for looking into this.

> I'm not sure that all of these parameters are preserved after serialization, have you tested if handleInvalid and defaultValue are the same after deserializing ?

Good question. I think I tested it only in real life and forgot to add such case in the unit test. I have added it to the test now in the open PR: https://github.com/combust/mleap/pull/624/commits/49cd0139df7a41f31b81d5b3b91f3a3526e78659

> I don't see how the python StringMap can be initialized correctly... https://github.com/combust/mleap/blob/d276c1c800277a1edcf30dda6e8559c7413b21f8/mleap-spark-extension/src/main/scala/org/apache/spark/ml/mleap/feature/StringMap.scala

For some reason that link is to an old commit? Note that handleInvalid and defaultValue are part of StringMapModel, see https://github.com/combust/mleap/blob/master/mleap-core/src/main/scala/ml/combust/mleap/core/feature/StringMapModel.scala.

(if you check StringMapModel in the same commit as your link, they are not there yet https://github.com/combust/mleap/blob/d276c1c800277a1edcf30dda6e8559c7413b21f8/mleap-core/src/main/scala/ml/combust/mleap/core/feature/StringMapModel.scala)

> I also see the deserialization test is inactive

Oh, but the comment already said that it works and is just about test setup in Travis. It was initially disabled for Travis because of a problem related to multi-scala build. You can see that the Travis build has succeeded for commit 24a6ad5 in https://github.com/combust/mleap/pull/624, which has a fix for the class path setup. The last commit was just updating with master, but somehow Travis failed to start a build entirely, meh. But as I said you can see the proof that test_serialize_to_bundle works in that PR. It tests that 'a' is mapped to 1.0 by the deserialized StringMap.

Now with my new commit Travis build seems to have started. I hope there are no random infrastructure failures this time and it passes.

I haven't heard about @ancasarb in a long time. I wonder how to get that PR merged.

lucagiovagnoli: I think you might be right, in that previous blob, loading 'handleInvalid' here https://github.com/combust/mleap/blob/6d2725aa9dff9bbb0f7deec465f49edf0ec78107/mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/StringMapOp.scala#L33 was missing but I see you've added it

PR #572, opened 2019-09-24
Validation & API improvements to string_map pyspark wrapper

juhoautio:
- Remove separate `from_dataframe` method. Allow passing a `DataFrame` as an alternative to a `dict` in the same constructor.
- Validate the args as much as possible before passing to JVM, because otherwise user gets some hard to interpret Py4j error

This code has been unit-tested. See https://github.com/combust/mleap/pull/571 for details. The test tooling for pyspark is not yet in a shape to be merged in.

### DIFF
--- a/python/mleap/pyspark/feature/string_map.py
+++ b/python/mleap/pyspark/feature/string_map.py
@@ -1,17 +1,35 @@
 from pyspark.ml.util import JavaMLReadable, JavaMLWritable, _jvm
 from pyspark.ml.wrapper import JavaTransformer
 from pyspark.ml.param.shared import HasInputCol, HasOutputCol
+from pyspark.sql import DataFrame
 
 
 class StringMap(JavaTransformer, HasInputCol, HasOutputCol, JavaMLReadable, JavaMLWritable):
     def __init__(self, labels, inputCol=None, outputCol=None, handleInvalid='error', defaultValue=0.0):
         """
         __init__(self, labels, inputCol=None, outputCol=None, handleInvalid='error', defaultValue=0.0)
-        labels is a dict {string: double}
+        labels must be a dict {string: double} or a spark DataFrame with columns inputCol & outputCol
         handleInvalid: how to handle missing labels: 'error' (throw an error), or 'keep' (map to the default value)
         """
-        assert handleInvalid in ['error', 'keep'], 'Invalid value for handleInvalid: {}'.format(handleInvalid)
         super(StringMap, self).__init__()
+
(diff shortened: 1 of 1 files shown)


python/mleap/pyspark/feature/string_map.py
```diff
@@ -1,17 +1,35 @@
...
-        labels is a dict {string: double}
+        labels must be a dict {string: double} or a spark DataFrame with columns inputCol & outputCol
         handleInvalid: how to handle missing labels: 'error' (throw an error), or 'keep' (map to the default value)
         """
-        assert handleInvalid in ['error', 'keep'], 'Invalid value for handleInvalid: {}'.format(handleInvalid)
         super(StringMap, self).__init__()
+
+        if isinstance(labels, DataFrame):
+            assert inputCol, 'inputCol is required when labels is a DataFrame'
+            assert outputCol, 'outputCol is required when labels is a DataFrame'
```
ancasarb: I think the message needs updating, they are required always, as they're used on lines 45/46 no?

ancasarb: we should probably move all the validations in validate_args() method, as well.

juhoautio: I think it's possible to leave `inputCol` & `outputCol` as `None` when creating the `StringMap`. Then they can be afterwards set, like this:

```python
string_map = StringMap(labels)
string_map.setInputCol(inputCol)
string_map.setOutputCol(outputCol)
```

However, for the conversion from DataFrame to dict they're absolutely needed. 

It seems that I copied this API from the OneHotEncoder wrapper that used to exist in mleap. And I think OneHotEncoder had been copied from spark's own python wrapper for `StringIndexer`, although that mimicking didn't go as far as copying the `StringIndexer.setParams`.

To me it seems like it would be maybe better to always require them in constructor for `StringMap`. That would make it clear what are mandatory args before the StringMap transformer can be actually used? Or is there some use case where one would want to test the transformer's plain function without specifying input / output col names?

IMHO a viable choice is to leave the code as it is in this PR.

juhoautio: > we should probably move all the validations in validate_args() method, as well, do the validations first, and then do the instance checks on dataframe

This block is purely about the DataFrame -> dict conversion case. That's why it's not in `validate_args`.

juhoautio:
Rather merge this change as part of https://github.com/combust/mleap/pull/571, which also has unit tests to validate the code

PR #593, opened 2019-11-05
Fix typo in README. Add a link to PyPI page.

juhoautio:
Btw, PyPI only seems to have this old version:

> mleap 0.8.1
> Last released: Oct 9, 2017

While this in git:

https://github.com/combust/mleap/blob/31db771ae26ec7b1e75864bb29361deaaa79c3f8/python/mleap/version.py#L19

Are you going to publish the latest version to PyPI?

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,11 +103,10 @@ $ bin/spark-shell --packages ml.combust.mleap:mleap-spark_2.11:0.15.0
 
 ### PySpark Integration
 
-Install MLeap from pypy
+Install MLeap from [PyPI](https://pypi.org/project/mleap/)
 ```bash
 $ pip install mleap
 ```
-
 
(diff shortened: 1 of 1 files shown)


ancasarb:
Thanks for this! Just posted the following on Gitter too 

```
@here This weekend we released MLeap version 0.15.0 (the JVM packages). You can take a look at the release notes here https://github.com/combust/mleap/blob/master/RELEASE_NOTES.md. Going forward, I will keep track of PRs merged and update the release notes as we go along so that they are even more accurate. My plan is to update the Gitbook documentation and release the Python package this coming weekend, I will post an update once that’s done. Following this, I will start looking at the PR for supporting custom transformers in Python, with a release following shortly after that’s merged.```

PR #600, opened 2019-11-19
Adding Batch leap frame and a sample batch tf transformer

sushrutikhar:
Currently in mleap we only have default leapframe which applies transformation to the dataset row by row. However, as TF does support predictions over a batch of requests and is internally optimised for that, we can leverage the benefits in mleap using a batch leap frame. This increases the throughput and decreases the latencies as opposed to a sequential processing.
A BatchTransformer will take Seq[Row] as input and return back the transformed and enriched putput as Seq[Row] 
A sample BatchTensorflowTransformer is added in this PR


sushrutikhar:
@hollinwilkins @ancasarb

sushrutikhar:
hey @ancasarb did you get a chance to have a look at the PR?

lucagiovagnoli:
This looks interesting.
@sushrutikhar do you think our benchmark analysis in https://github.com/combust/mleap/issues/631 (between xgboost4j and mleap) could be related? The factor 2x might be a pattern between the two ?

sushrutikhar:
> This looks interesting.
> @sushrutikhar do you think our benchmark analysis in #631 (between xgboost4j and mleap) could be related? The factor 2x might be a pattern between the two ?

@lucagiovagnoli the gain we saw is mainly because the default leap frame is not utilising underlying libraries's ability to do parallel processing. Looks like in your case the xgboost library being used is itself having performance issues. however, as an exercise we can try using the parallel leap frame introduced in this PR for xgbooost as well and see if that gives more performance gain over and above to the changes you proposed in your PR

PR #605, opened 2019-11-26
Fix xgboost sparse vector support

talalryz:
**What is the bug?**
mleap xgboost does not support sparse vector

**How can the bug be found?**
I added some sparse rows to the testing data in `xgboost_training.csv`. This causes tests to fail with predictions from mleap not matching spark predictions.
I first noticed this bug in https://github.com/combust/mleap/issues/596

**How will the bug be fixed?**
Create DMatrix directly from sparseTensor and sparseVector, instead of converting to denseVector and denseTensor first.

**What does this PR do?**
1. Implements the fix above.
2. Also fixes a bug in XGBoostClassifier where predict doesn't pass treeLimit 

**Tests**
The regressor test now passes with updated dataset that includes sparse Data. 
For the classifier, i added a new test that uses this mixed dataset. This also passes.

** WHAT THIS TEST DOES NOT FIX**
These changes have been tested with mleap v0.11.0, and should work until 0.14. ~~These changes do not fix issues with v0.15.0 which uses xgboost v0.90. XGBoost v0.90 no longer supports using Float.NaN as missing values for sparse vectors, and I have not found a way to make mleap predictions match spark predictions.
There are issues that lie within xgboost itself, and we should be able to resolve after this approved PR to xgboost is merged: https://github.com/dmlc/xgboost/pull/4805~~

EDIT: I updated the tests to match the xgboost specification for handling missing values: https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#dealing-with-missing-values

This branch will work with xgboost v0.90.


mleap-xgboost-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostClassificationModelParitySpec.scala
```diff
@@ -106,7 +129,6 @@ class XGBoostClassificationModelParitySpec extends FunSpec
             }
             assert(Math.abs(v2 - v1) < 0.0000001)
```
lucagiovagnoli: Magic number in code :S
Could this maybe be 
`SEVEN_SIGNIFICANT_DIGITS = 0.0000001`

mleap-xgboost-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostClassificationModelParitySpec.scala
```diff
@@ -136,32 +159,39 @@ class XGBoostClassificationModelParitySpec extends FunSpec
...
     val sparkDataset = sparkTransformer.transform(dataset)
     val mleapSchema = TypeConverters.sparkSchemaToMleapSchema(dataset)
 
     val data = dataset.collect().map {
       r => Row(r.toSeq: _*)
     }
     val frame = DefaultLeapFrame(mleapSchema, data)
-    val mleapT = mleapTransformer(sparkTransformer)
+    val mleapT = mleapTransformer(sparkTransformer, dataset, bundleCache)
     val mleapDataset = mleapT.transform(frame).get
```
lucagiovagnoli: this should ideally be one responsibility

lucagiovagnoli:
@talalryz Do you know why tests are not passing ?

mleap-xgboost-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostClassificationModelParitySpec.scala
```diff
@@ -46,30 +48,51 @@ class XGBoostClassificationModelParitySpec extends FunSpec
...
 
-  val dataset: DataFrame = {
+  val denseDataset: DataFrame = {
     SparkParityBase.dataset(spark).select("fico_score_group_fnl", "dti").
       filter(col("fico_score_group_fnl") === "500 - 550" ||
         col("fico_score_group_fnl") === "600 - 650")
   }
 
-  val sparkTransformer: Transformer = {
+  val mixedDataset: DataFrame = {
```
lucagiovagnoli: This is supposed to be a Sparse dataset at some point along the way right ? Due to `implicit class VectorOps` being implicit I am not sure where that happens

mleap-xgboost-spark/src/test/resources/datasources/xgboost_training.csv
```diff
@@ -986,15 +986,15 @@
...
-14.26,40.92,1022.07,73.96,460.38
-21.38,58.33,1013.05,72.75,446.8
-15.71,44.06,1018.34,71.69,465.05
-5.78,40.62,1016.55,84.98,484.13
-6.77,39.81,1017.01,87.68,488.27
-23.84,49.21,1013.85,50.36,447.09
-21.17,58.16,1017.16,68.11,452.02
-19.94,58.96,1014.16,66.27,455.55
-8.73,41.92,1029.41,89.72,480.99
+21.91,0,0,0,445.04
```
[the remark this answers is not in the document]
talalryz: They are representing missing values. The spark vector assembler will form sparse vectors corresponding to these rows. The test setup is a pipeline of,

> features -> vector assembler -> xgboost model 

Having these rows with missing values means that the  Mleap xgboost models will actually have to deal with SparseVectors ( or SparseTensors)

mleap-xgboost-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostClassificationModelParitySpec.scala
talalryz: > @talalryz Do you know why tests are not passing ?

the tests are not passing because xgboost 0.90 does not allow Float.NaN as missing value for  sparse Vectors.  

> ERROR DataBatch: java.lang.RuntimeException: you can only specify missing value as 0.0 (the currently set value NaN) when you have SparseVector or Empty vector as your feature format

This is expected as this branch does not fix the issues introduced by the recent upgrade to xgboost 0.90.

mleap-xgboost-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostClassificationModelParitySpec.scala
talalryz: I called it a mixed Dataset because it has both sparse and dense rows. I can rename it sparseDataset to be more clear

lucagiovagnoli: Mmh, I don't know if a dataset having both sparse rows and "full" rows would still sparse or dense. Is the glass half empty or half full?
But I think for readability it's good to have a sparkeDataset and a denseDataset, so the tests are clearer

talalryz:
I updated the tests in this branch to match the xgboost specification for dealing with missing values: https://xgboost.readthedocs.io/en/latest/jvm/xgboost4j_spark_tutorial.html#dealing-with-missing-values

All the tests now pass. 

@ancasarb

mleap-xgboost-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostClassificationModelParitySpec.scala
talalryz: updated the branch so that all tests pass

mleap-xgboost-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostClassificationModelParitySpec.scala
talalryz: updated naming to sparseDataset

mleap-xgboost-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostRegressionModelParitySpec.scala
```diff
@@ -36,6 +36,7 @@ class XGBoostRegressionModelParitySpec extends FunSpec
   private val xgboostParams: Map[String, Any] = Map(
     "eta" -> 0.3,
     "max_depth" -> 2,
+    "missing" -> 0.0f,
```
lucagiovagnoli: Does this mean that missing are substituted with zeros ?

lucagiovagnoli: I'm wondering, why not putting nulls here ?

talalryz: **Short Answer:**

1. This means that zeros will be assumed to be missing values.

2. We can't set it as anything other than 0.0 because xgboost throws the following error if we do that:

`java.lang.RuntimeException: you can only specify missing value as 0.0 (the currently set value NaN) when you have SparseVector or Empty vector as your feature format
`

**Detailed Answer:**

The logic behind not allowing any value other than 0.0  is that Sparks VectorAssembler always treats 0 as a missing value, so any SparseVector created by the Spark VectorAssembler will omit 0.0 treating it as missing. 
XGboost devs decided that to ensure compatibility with Sparks VectorAssembler they would enforce setting `missing` as 0.0 in their models whenever a SparseVector was used., otherwise; there would be incompatibility between DenseVector and SparseVector 

There is more info here: https://github.com/dmlc/xgboost/pull/4349/files

PR #615, opened 2019-12-13
[Feature] Python Support for Multinomial Logistic Regression

voganrc:
Hi There,

I need mleap to serialize a multi-class logistic regression model that I trained using scikit-learn. I noticed, however, that the number of classes is hard-coded as `2`:

https://github.com/combust/mleap/blob/beb4fd248e020282f10778c91d68a3518e981dcb/python/mleap/sklearn/logistic.py#L72

So I took the liberty of implementing that `TODO` myself.

In the process, I believe that I also found a bug in the way mleap serializes numpy arrays. This line:

https://github.com/combust/mleap/blob/beb4fd248e020282f10778c91d68a3518e981dcb/python/mleap/bundle/serialize.py#L94

generates the following "shape" dictionary to represent a 4x1 np array:

```json
"shape": {
    "dimensions": [
       {
          "size": [
             4,
             1
          ],
          "name": ""
       }
    ]
 }
```

But I updated the `MLeapSerializer` to produce the following instead:

```json
"shape": {
    "dimensions": [
       {
          "size": 4,
          "name": ""
       },
       {
          "size": 1,
          "name": ""
       }
    ]
 }
 ```

which is the same format I found after serializing a spark.ml logistic regression model.

I couldn't find any integration tests for the python mleap code, so it's hard for me to know whether this change breaks anything in the scala side. But I tested this manually by deserializing two different bundles into the mleap runtime. The `master` branch serialization format causes the following error:

```
spray.json.DeserializationException: Expected Int as JsNumber, but got [4,1]
at spray.json.package$.deserializationError(package.scala:23)
at spray.json.ProductFormats$class.fromField(ProductFormats.scala:62)
at spray.json.DefaultJsonProtocol$.fromField(DefaultJsonProtocol.scala:30)
at spray.json.ProductFormatsInstances$$anon$2.read(ProductFormatsInstances.scala:56)
at spray.json.ProductFormatsInstances$$anon$2.read(ProductFormatsInstances.scala:46)
at spray.json.JsValue.convertTo(JsValue.scala:31)
at ml.combust.bundle.serializer.JsonFormatModelSerializer.read(ModelSerializer.scala:52)
at ml.combust.bundle.serializer.ModelSerializer$$anonfun$readWithModel$1.apply(ModelSerializer.scala:103)
at ml.combust.bundle.serializer.ModelSerializer$$anonfun$readWithModel$1.apply(ModelSerializer.scala:103)
at scala.util.Try$.apply(Try.scala:192)
...
```

Wheras a bundle generated on my updated branch works fine.

Thanks!
-Ryan

### CODE REVIEW

File: python/tests/sklearn/base_tests.py [hunk 1]
```diff
@@ -155,6 +174,60 @@ def test_logistic_regression_deserializer(self):
         self.assertEqual(res_a[1], res_b[1])
         self.assertEqual(res_a[2], res_b[2])
 
+    def test_multinomial_logistic_regression_serializer(self):
+
+        logistic_regression = LogisticRegression(fit_intercept=True)
+        logistic_regression.mlinit(
+            input_features='features',
```
lucagiovagnoli: should this be `input_features='a'`?

python/tests/sklearn/base_tests.py
```diff
@@ -31,9 +31,28 @@
 from mleap.sklearn.preprocessing.data import FeatureExtractor, Binarizer
 
 
+def to_quartile(rand_num):
```
lucagiovagnoli: One thing, this is very specific to a standard normal distribution (mean 0 and variance 1?). Could this be reflected in the naming?

python/tests/sklearn/base_tests.py
```diff
@@ -155,6 +174,60 @@ def test_logistic_regression_deserializer(self):
...
+        # assert 4x0 intercept vector
+        self.assertEqual(len(model['attributes']['intercept_vector']['double']), 4)
+        self.assertEqual(len(model['attributes']['intercept_vector']['shape']['dimensions']), 1)
+        self.assertEqual(model['attributes']['intercept_vector']['shape']['dimensions'][0]['size'], 4)
+
+    def test_multinomial_logistic_regression_deserializer(self):
+
+        logistic_regression = LogisticRegression(fit_intercept=True)
+        logistic_regression.mlinit(
+            input_features='features',
```
lucagiovagnoli: this one too? Sorry, should have pointed it out earlier

voganrc: All good, just noticed it myself

python/mleap/sklearn/logistic.py
```diff
@@ -102,6 +110,8 @@ def deserialize_from_bundle(self, transformer, node_path, node_name):
             transformer.fit_intercept = False
 
         transformer.coef_ = np.array([transformer.coef_])
```
ancasarb: just double checking, does this need to happen for both the binary and the multinomial logistic regression? or could it be the else statement in the if below?

python/mleap/sklearn/logistic.py
```diff
@@ -88,7 +94,9 @@ def deserialize_from_bundle(self, transformer, node_path, node_name):
 
         attributes_map = {
             'coefficients': 'coef_',
-            'intercept': 'intercept_'
+            'coefficient_matrix': 'coef_',
+            'intercept': 'intercept_',
+            'intercept_vector': 'intercept_',
```
ancasarb: probably we should have the if statement for the number of classes here as well, what do you think?

python/mleap/sklearn/logistic.py
voganrc: I turned this into an `np.reshape` operation to make the shape of `transformer.coef_` more explicit for binary classification

python/mleap/sklearn/logistic.py
voganrc: You need to deserialize the transformer to know its number of classes. And you can't deserialize without already having this attribute map.

Issue #635, opened 2020-01-31
[XGBoostClassification] Runtime predictRaw doesn't match dmlc-xgboost


lucagiovagnoli:
I think we found a bug in MLeap [here](https://github.com/combust/mleap/blob/master/mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModel.scala#L45) where `predictRaw` doesn't match what dmlc-xgboost does.

Basically it should return (-m, m) rather than (1-m, m)
The latter is the correct result for `predictProbabilities` [here](https://github.com/combust/mleap/blob/master/mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModel.scala#L40), could it be a copy-paste leftover ?

Fix is to replicate what xgboost does [here](https://github.com/dmlc/xgboost/blob/37fdfa03f86bd4bfb8b954e7bb2d4086bd43ce8b/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala#L444)

lucagiovagnoli:
Fixed in #637

PR #637, opened 2020-02-08
Luca/fix raw predict issue 635

lucagiovagnoli:
**Code change**
This is a one line change that fixes issue #635 where rawPredict should be `(-m, m)`.
(also found another small bug where tree_limit wasn't being serialized in the `store`)

**Added tests**
I noticed there were no tests for mleap-xgboost-runtime so I added a bunch of them using the same "agaricus" dataset from `dmlc-xgboost`:

1. Booster and mleap XGBoostClassifier results match (tests for `(-m, m)`)
2. XGBoostClassification works before and after serialization.
3. XGBoostRuntime schema assertions based on the NodeShape

@talalryz @voganrc @ancasarb


mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/testing/BundleSerializationUtils.scala
```diff
@@ -0,0 +1,35 @@
...
+import ml.combust.mleap.runtime.{MleapContext, frame}
+import ml.combust.mleap.runtime.frame.Transformer
+import resource.managed
+
+trait BundleSerializationUtils {
+
+  def serializeModelToMleapBundle(transformer: Transformer): File = {
+    import ml.combust.mleap.runtime.MleapSupport._
+
+    new File("/tmp/mleap/spark-parity").mkdirs()
```
ancasarb: perhaps "/tmp/mleap/xgboost-runtime-parity" as this doesn't test anything spark related?

project/Dependencies.scala
```diff
@@ -114,7 +115,7 @@ object Dependencies {
 
   val tensorflow = l ++= tensorflowDeps ++ Seq(Test.scalaTest)
 
-  val xgboostRuntime = l ++= Seq(xgboostDep) ++ Seq(Test.scalaTest)
+  val xgboostRuntime = l ++= Seq(xgboostDep) ++ Test.spark ++ Seq(Test.scalaTest)
```
ancasarb: are you sure Provided.spark isn't enough? it seems like that for the other parity tests?

project/MleapProject.scala
```diff
@@ -110,7 +110,9 @@ object MleapProject {
   lazy val xgboostRuntime = Project(
     id = "mleap-xgboost-runtime",
     base = file("mleap-xgboost-runtime"),
-    dependencies = Seq(runtime)
+    dependencies = Seq(
+      runtime,
+      sparkTestkit % "test")
```
ancasarb: where do we use sparkTestkit?

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/testing/CachedDatasetUtils.scala
```diff
@@ -0,0 +1,50 @@
...
+  lazy val leapFrameLibSVMtest: DefaultLeapFrame = leapFrameFromLibSVMFile(TestDataFilePath)
+  lazy val numFeatures: Int =
+    leapFrameLibSVMtrain.schema.getField("features").get.dataType.asInstanceOf[TensorType].dimensions.get.head
+
+  private def leapFrameFromLibSVMFile(filePath: String): DefaultLeapFrame = {
+
+    // Use Spark utils to load libsvm from disk
+    val spark = SparkSession.builder()
+      .master("local[2]")
+      .appName("XGBoostRuntimeClassificationModelParitySpec")
```
ancasarb: is this used for both tests?

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/bundle/ops/XGBoostClassificationOp.scala
```diff
@@ -23,8 +23,10 @@ class XGBoostClassificationOp extends MleapOp[XGBoostClassification, XGBoostClas
                       (implicit context: BundleContext[MleapContext]): Model = {
       val out = Files.newOutputStream(context.file("xgboost.model"))
       obj.booster.saveModel(out)
-      model.withValue("num_features", Value.int(obj.numFeatures)).
-        withValue("num_classes", Value.int(obj.numClasses))
+      model
+        .withValue("num_features", Value.int(obj.numFeatures))
+        .withValue("num_classes", Value.int(obj.numClasses))
+        .withValue("tree_limit", Value.int(obj.treeLimit))
```
talalryz: Was treeLimit not being saved earlier, and this was a bug. If so, I think we should report it as a bug, and mention that bug here. It will help folks facing this issue!

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/testing/BoosterUtils.scala
```diff
@@ -0,0 +1,18 @@
+package ml.combust.mleap.xgboost.runtime.testing
+
+import ml.dmlc.xgboost4j.scala.{Booster, DMatrix, XGBoost}
+
+trait BoosterUtils {
+
+  final val xgboostParams: Map[String, Any] = Map(
+    "eta" -> 0.3,
+    "max_depth" -> 2,
+    "objective" -> "binary:logistic",
```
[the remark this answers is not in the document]
lucagiovagnoli: Yes, I think this should be already tested in the XGBoost library, no need to cross the boundary.

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/bundle/ops/XGBoostClassificationOp.scala
lucagiovagnoli: Serializing a model using this Op makes it unusable because the `load()` expects a `tree_limit` and it blows up when it's not there. But I am not sure if anyone stores a Runtime model? 

In any case, I opened an issue here https://github.com/combust/mleap/issues/639

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/testing/BundleSerializationUtils.scala
lucagiovagnoli: Oh yes, leftover from copy-pasting from the spark tests. Let me update it

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/testing/CachedDatasetUtils.scala
lucagiovagnoli: Yes. I'll change it to this.getClass.getName

project/Dependencies.scala
lucagiovagnoli: So I actually meant to get your input on these dependency changes. 
Both Provided.spark or Test.spark achieve the same thing here because I am using spark things only in tests.

Now, I think that putting those dependencies with a `test` <scope> makes the build faster, since the dependencies are only brought in when running tests, not when building. I read online:

- test: 
> Dependencies with maven dependency scope test are not needed to build and run the project. They are needed to compile and run the unit tests.

- provided:
> Maven dependency scope provided is used during build and test the project. 

so I assume a test scope is slightly lighter than provided? Again, the xgboost-runtime code has no dependencies on spark stuff, but its tests do

project/MleapProject.scala
lucagiovagnoli: If I don't bring that in, I cannot use:
`import org.apache.spark.sql.mleap.TypeConverters`

 Do you think it is ok to use that in tests? (%"test" should not be included in the build)

lucagiovagnoli:
Thanks all for reviewing! I added a multinomial classifier test and fixed all copy-paste leftovers. Let me know what you think about using spark % test for these tests.

lucagiovagnoli:
it seems that my new tests passed successfully (https://travis-ci.org/combust/mleap/jobs/649255264#L9688) but the python ones failed. I'll try rerun these

lucagiovagnoli:
I think this broke after the recent update of virtualenv to 20.0.0
See https://github.com/pypa/virtualenv/issues/1551

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModelParitySpec.scala
```diff
@@ -0,0 +1,128 @@
...
+
+    leapFrameLibSVMtest.dataset.foreach {
+      r=>
+        val mleapResult = mleapTransformer.transform(DefaultLeapFrame(mleapSchema.get, Seq(r))).get
+
+        val mleapPredictionColIndex = mleapResult.schema.indexOf("prediction").get
+        val mleapRawPredictionColIndex = mleapResult.schema.indexOf("raw_prediction").get
+        val mleapProbabilityColIndex = mleapResult.schema.indexOf("probability").get
+
+        val singleRowDMatrix = r(1).asInstanceOf[SparseTensor[Double]].asXGB
```
voganrc: Is the `1` here the index of the assembled features?

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModelParitySpec.scala
```diff
@@ -0,0 +1,128 @@
...
+        ))
+    }
+  }
+
+  it("Results between the XGBoost4j booster and the MLeap Transformer are the same") {
+    val booster = trainBooster(xgboostParams, denseDataset)
+    val xgboostTransformer = createBoosterClassifier(trainBooster(xgboostParams, denseDataset))
+
+    val mleapBundle = serializeModelToMleapBundle(xgboostTransformer)
+    val deserializedTransformer: Transformer = loadMleapTransformerFromBundle(mleapBundle)
```
voganrc: Why do we bother serializing/deserializing here if the test below asserts pre/post serialization consistency?

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModelParitySpec.scala
```diff
@@ -0,0 +1,128 @@
...
+        assert (almostEqualSequences(
+          Seq(boosterRawPrediction.values),
+          Seq(mleapResult.dataset.head.getTensor[Double](mleapRawPredictionColIndex).toArray)
+        ))
+    }
+  }
+
+  it("Results between the XGBoost4j booster and the MLeap Transformer are the same") {
+    val booster = trainBooster(xgboostParams, denseDataset)
+    val xgboostTransformer = createBoosterClassifier(trainBooster(xgboostParams, denseDataset))
```
voganrc: So the transformer is created via another `trainBooster` call? How do we guarantee that it ends up building the same tree? Are the xgboostParams seeded?

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModelParitySpec.scala
voganrc: If serialization breaks, then both of these tests fail, but I think it would be better for only one to fail

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModelParitySpec.scala
```diff
@@ -0,0 +1,128 @@
...
+
+    val deserializedTransformer: Transformer = loadMleapTransformerFromBundle(mleapBundle)
+    val deserializedModelResult = deserializedTransformer.transform(leapFrameLibSVMtrain).get
+
+    assert(preSerializationResult.get.dataset == deserializedModelResult.dataset)
+  }
+
+  it("Test XGBoostMultinomialClassificationModel results are the same pre and post serialization") {
+    val booster = trainBooster(xgboostParams, denseDataset)
+    val xgboostTransformer = createMultinomialBoosterClassifier(booster)
```
voganrc: Is there a test for `MultinomialBoosterClassifier` parity with XGBoost4J anywhere?

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/testing/FloatingPointApproximations.scala
```diff
@@ -0,0 +1,24 @@
...
+
+trait FloatingPointApproximations {
+
+  val DefaultMinimumPrecision = 1e-7
+
+  def almostEqual(x: Double, y: Double, precision: Double = DefaultMinimumPrecision): Boolean = {
+    if ((x - y).abs < precision) true else false
+  }
+
+  def almostEqualSequences(
```
voganrc: Is it really necessary to implement this ourselves? Is there no implementation in a library somewhere?

lucagiovagnoli: We looked into it last quarter and haven't found anything. This is the same code we currently use internally to check float equality to the 1e-7

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModelParitySpec.scala
lucagiovagnoli: Originally I was re-using the booster object but then I thought that it must probably create the same tree by construction. It seems that it does - otherwise this test would/will fail

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModelParitySpec.scala
lucagiovagnoli: True, I'll remove that from here

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModelParitySpec.scala
lucagiovagnoli: Yes, I'll use a constant name to be clearer
* row(0) is the label
* row(1) is "features"

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModelParitySpec.scala
lucagiovagnoli: I added one using a mutli-label Iris dataset

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/testing/CachedDatasetUtils.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  private def leapFrameFromLibSVMFile(filePath: String): DefaultLeapFrame = {
+
+    // Use Spark utils to load libsvm from disk
+    val spark = SparkSession.builder()
+      .master("local[2]")
+      .appName(s"${this.getClass.getName}")
+      .getOrCreate()
+
+    // This is the dataset used by dmls-XGBoost https://github.com/dmlc/xgboost/blob/master/demo/data/agaricus.txt.train
```
talalryz: typo, dmls --> dmlc

project/MleapProject.scala
ancasarb: I think that comes from mleap-spark-base, but spark-testkit should be ok too.

project/Dependencies.scala
ancasarb: That makes sense, thank you!

PR #641, opened 2020-02-13
Fix python test failure using virtualenv<20.0.0

### LINKED ISSUE #640: Python tests broken by virtualenv upgrade to 20.0.0 on Travis
My other branch https://github.com/combust/mleap/pull/637 tests started failing today on Travis (but it wasn't failing 5 days ago).

This seems to be a very recent bug in virtualenv opened just 3 days ago, more details here:
https://github.com/pypa/virtualenv/issues/1551

lucagiovagnoli:
Fixes #640  and #642  and **improves tests running time from ~24min down to ~15min.**

1. Fixes recent virtualenv==20.0.0 bug that broke Travis builds 3 days ago (https://github.com/pypa/virtualenv/issues/1551).
2. Runs all python tests with all virtualenvs (for real! We were not running all of py27, py36, py37).
3. Python 2.7 is officially deprecated. **Let me know if you'd rather remove this**
4. Runs tests using Travis stages, so that we can run them in parallel and that we can actually use the correct system python version == tox python version. The problem seems to be that `$TRAVIS_PYTHON_VERSION` is unset (see #642). It looks like other projects on Github are using Travis stages to install the correct python version in each VM that is compatible to a specific tox environment.

One thing to note, unfortunately, due to some python tests relying on these jars: https://github.com/combust/mleap/blob/master/mleap-spark-extension/build.sbt#L9
I am forced to run `sbt +compile` before python tests in all stages (running in different VMs) and this makes them all slower because they are all compiling the entire project.

@talalryz @lydian @voganrc  for review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,20 +5,50 @@ dist: trusty
 services:
   - docker
 
-language: scala
 scala:
   - 2.11.8
+jdk:
+  - oraclejdk8  # sbt +compile fails on jdk11 (xenial's default)
 
-python:
-  - "2.7"
-  - "3.6"
-  - "3.7"
+jobs:
+  include:
+    - stage: "Mleap tests"
+      name: "Scala Tests"
+      language: scala
+      script:
+        - travis/travis.sh
 
-install:
-    -   pip install --user tox
-script:
-  - travis/travis.sh
-  - tox -c python/tox.ini -e $(echo py$TRAVIS_PYTHON_VERSION | tr -d .) -v
+    - name: "Python 2.7 tests"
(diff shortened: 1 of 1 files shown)


lucagiovagnoli:
scala tests and py36 passes.
`sbt +compile` fails in py37 tests. I wonder if that's because I changed the linux distro to xenial (as py37 is only available for xenial: https://travis-ci.community/t/unable-to-download-python-3-7-archive-on-travis-ci/639/2).

I could also try py35 and py38. We're not running on py37 right now anyways.

PR #645, opened 2020-02-15
xgboost Predictor performant Op

### LINKED ISSUE #631: XGBoost Performance Issues
Hello,

I ran some JMH benchmarks that show MLeap to be significantly slower than other libraries for evaluating XGBoost models.

image

Here you can see throughput (ops / sec) as a function of library and batch size, where:

xgboost4j = https://github.com/dmlc/xgboost/tree/master/jvm-packages
xgboost-predictor-java = https://github.com/komiya-atsushi/xgboost-predictor-java
yelp-xgboost = https://github.com/Yelp/xgboost-predictor-java
mleap = https://github.com/combust/mleap

Given that Mleap makes use of `xgboost4j-spark` does anyone know why it would have half the throughput of `xgboost4j`? Also, is there a reason why `mleap` does not observe constant throughput scaling like `xgboost4j` does?

Thanks!
-Ryan

lucagiovagnoli:
I added a new MLeapOp to load xgboost models as Predictor objects (more performant than xgboost4j). Honorable mentions to @hollinwilkins who introduced some of this code in 2017(see https://github.com/combust/mleap/pull/259)

In order to use this, it's enough to use a reference.conf file like the following:
```
ml.combust.mleap.xgboost.ops = [
  "ml.combust.mleap.xgboost.runtime.bundle.ops.XGBoostPredictorClassificationOp",
  "ml.combust.mleap.xgboost.runtime.bundle.ops.XGBoostRegressionOp"
]
```

and add this to your project's pom file:
```
<!-- Append our reference.conf into MLeap's reference.conf so our Ops are registered -->
--
  | <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
  | <resource>reference.conf</resource>
  | </transformer>
```

This fixes #631 (for review @talalryz @ancasarb )


lucagiovagnoli:
I see in logs `Using Scala 2.12.8` so these have failed cause they are using the wrong scala version.

I think travis wants both language and scala version to be set at the same hierarchical level in the travis file:
```
language: scala
scala:
    - 2.11.8
```

I fixed it here and rebased already: https://github.com/combust/mleap/pull/637/commits/7dcd9bd842b63230970265372e30e05a03e5d0a7

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModel.scala
```diff
@@ -0,0 +1,103 @@
...
+
+case class XGBoostPredictorClassificationModel(impl: XGBoostPredictorClassificationModelBase) extends ProbabilisticClassificationModel {
+  override val numClasses: Int = impl.numClasses
+  override val numFeatures: Int = impl.numFeatures
+  def treeLimit: Int = impl.treeLimit
+
+  def predictor: Predictor = impl.predictor
+
+  def binaryClassificationModel: XGBoostPredictorBinaryClassificationModel = impl.asInstanceOf[XGBoostPredictorBinaryClassificationModel]
+  def multinomialClassificationModel: XGBoostMultinomialClassificationModel = impl.asInstanceOf[XGBoostMultinomialClassificationModel]
```
CatTail:
```
  def multinomialClassificationModel: XGBoostPredictorMultinomialClassificationModel = impl.asInstanceOf[XGBoostPredictorMultinomialClassificationModel]
```

lucagiovagnoli: Uh, good catch!

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModel.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  def predict(data: FVec): Double = impl.predict(data)
+
+  def predictLeaf(features: Vector): Seq[Double] = impl.predictLeaf(features)
```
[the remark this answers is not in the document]
lucagiovagnoli: mmh I'm not sure that I understand: 
1. `XGBoostPredictorClassificationModel` extends `ProbabilisticClassificationModel` (not `XGBoostPredictorClassificationModelBase`) but holds an internal `impl`[ementation] object that extends `XGBoostPredictorClassificationModelBase`.

2. And we are calling the trait XGBoostPredictorClassificationModelBase `predictLeaf` by calling such method on the `impl` object

lucagiovagnoli:
Tests are fixed. As I suspected it had to do with scala 2.12

According to this line https://github.com/combust/mleap/blob/master/project/Dependencies.scala#L70 there's no dmlc xgboost for scala 2.12
`xgboostRuntime` mistakenly added in `aggregatedProjects` in the 'MLeapProject' file was running mleap-xgboost-runtime tests using scala 2.12

ancasarb:
@lucagiovagnoli could you please pull the latest master here, thank you!

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModel.scala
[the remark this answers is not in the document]
talalryz: i understand! I'll drop the issue

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassification.scala
```diff
@@ -0,0 +1,31 @@
...
+  override val exec: UserDefinedFunction = {
+
+    // Since the Predictor is our performant implementation, we only compute probability for performance reasons.
+    val probability = shape.getOutput("probability").map {
+      _ => (data: FVec) => Some(model.predictProbabilities(data): Tensor[Double])
+    }.getOrElse((_: FVec) => None)
+
+    val f = (features: Tensor[Double]) => {
+      val data: FVec = features.asXGBPredictor
+      val rowData = Seq(probability(data).get)
```
tinaxi: What if `probability(data)` returns a `None`? Calling `get` on a `None` object will throw exception. Or are you intentionally letting it throw the error rather than checking it beforehand?

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModel.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  override def predictRaw(features: Vector): Vector = predictRaw(features.asXGBPredictor)
+  def predictRaw(data: FVec): Vector
+
+  override def predictProbabilities(features: Vector): Vector = predictProbabilities(features.asXGBPredictor)
+  def predictProbabilities(data: FVec): Vector
+
+  def predictLeaf(features: Vector): Seq[Double] = predictLeaf(features.asXGBPredictor)
+  def predictLeaf(data: FVec): Seq[Double] = predictor.predictLeaf(data, treeLimit).map(_.toDouble)
+
```
tinaxi: nit: extra empty line at the end.

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/struct/FVecFactory.scala
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  def fromDenseVector(denseVector: DenseVector): FVec = {
+    // Casting to floats, because doubles result in compounding differences from the c++ implementation
+    // https://github.com/komiya-atsushi/xgboost-predictor-java/issues/21
+    FVec.Transformer.fromArray(denseVector.values.map(_.toFloat), false)
+  }
+
+  def fromTensor(tensor: Tensor[Double]): FVec = FVecTensorImpl(tensor)
+
```
tinaxi: nit: extra empty line at the end.

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassification.scala
lucagiovagnoli: This file is copied from https://github.com/combust/mleap/blob/master/mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModel.scala and https://github.com/combust/mleap/blob/master/mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostRegressionModel.scala so I'm using the same proven code that it's already been used for the xgboost4j regressor and classifier.

It's good for exceptions to _throw_ when something unexpected happens.

tinaxi: Do you mean `XGBoostClassification` here? https://github.com/combust/mleap/blob/master/mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassification.scala#L36

It is filtering by `_.isDefined` first and then calling `_.get`, which will not cause any exceptions. However, directly calling `get` from an `Option` object is not ideal. The error log will just say something like `java.util.NoSuchElementException: None.get`, which is not informant for the users at all I think?

tinaxi:
LGTM!

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassification.scala
lucagiovagnoli: Oh yes, I removed the _isDefined_ filter because I don't think we should ever expect non-defined here.  "None" is really unexpected because XGBoost Predictor should always have "probability" defined, shouldn't it?  So it's good to fail fast without masking the exception to the user (I'd be happier to wrap any expected user-facing exceptions).

Even in the original code, if .isDefined was false, that value would be _.filtered_ out and the _Row_ would end up with the values all shifted left by the number of missing fields, which would break the output schema. I might be missing something here but I am not sure if isDefined makes sense. Unless it's just the norm to put it by default when accessing a scala Option object?

lucagiovagnoli:
Thanks for reviewing everyone!

@ancasarb this has now been rebased on master and tests pass. let me know what you think

lucagiovagnoli:
Last diff:
- fixes a copy-pasted duplicate test
- Better FVec Factory implementation so we don't subclass an external class. This allows people to choose the library they prefer (eg. https://github.com/komiya-atsushi/xgboost-predictor-java which I'm including here or any compatible fork like https://github.com/Yelp/xgboost-predictor-java)

lucagiovagnoli:
@tinaxi: your comment disappeared after I amended the commit. RE: I streamlined those imports and converted to `lang.Integer`

tinaxi:
LGTM!

lucagiovagnoli:
Hi @ancasarb ! We've been using this internally successfully for a while now, do you think we could merge it into master?

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/bundle/ops/XGBoostPredictorClassificationOp.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  override val Model: OpModel[MleapContext, XGBoostPredictorClassificationModel] = new OpModel[MleapContext, XGBoostPredictorClassificationModel] {
+  override val klazz: Class[XGBoostPredictorClassificationModel] = classOf[XGBoostPredictorClassificationModel]
+
+  override def opName: String = "xgboost.classifier"
+
+  @throws[RuntimeException]
+  override def store(model: Model, obj: XGBoostPredictorClassificationModel)
+                    (implicit context: BundleContext[MleapContext]): Model =
+    throw new RuntimeException("The XGBoostPredictor implementation does not support storing the model.")
```
ancasarb, line 27: Is it because the underlying library doesn't support this?

lucagiovagnoli: Yes correct. The library's only purpose is to load an xjboost4j model runtime to predict/run/infer with higher performance than the officail xgboost4j loader.

They achieve this by re-implementing the runtime code in Java which is faster than going through C++  (which is what xgboost4j does). Going to C++ is fine in general, but doesn't really scale on "big data" traffic, so I think it's good to provide both options in MLeap.

Some of the benchmarks are on their main page: https://github.com/komiya-atsushi/xgboost-predictor-java
but one of our teammates has also run independent benchmarks here: https://github.com/combust/mleap/issues/631

project/Common.scala
```diff
@@ -22,6 +22,7 @@ object Common {
     fork in Test := true,
     javaOptions in test += sys.env.getOrElse("JVM_OPTS", ""),
     resolvers += Resolver.mavenLocal,
+    resolvers += Resolver.jcenterRepo,
```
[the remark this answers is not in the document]
lucagiovagnoli: This is required cause that's where the predictor lives, it's how they suggest to import it on their main page: https://github.com/komiya-atsushi/xgboost-predictor-java (note that the *original* version is https://github.com/komiya-atsushi/xgboost-predictor-java, not https://github.com/h2oai/xgboost-predictor ;)

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/struct/FVecFactory.scala
```diff
@@ -0,0 +1,51 @@
...
+import scala.collection.JavaConversions.mapAsJavaMap
+
+
+object FVecFactory {
+  private def toJavaMap(map: Map[Int, Float]): util.Map[lang.Integer, lang.Float] = {
+    mapAsJavaMap(map).asInstanceOf[util.Map[lang.Integer, lang.Float]]
+  }
+
+  /**
+    *  NOTE: all these methods cast doubles to floats, because doubles result in compounding differences
```
lucagiovagnoli, line 18: @ancasarb reference to this comment regarding floats vs double bug

project/Dependencies.scala
```diff
@@ -68,6 +68,8 @@ object Dependencies {
       "com.typesafe.scala-logging" %% "scala-logging" % loggingVersion
     )
     val xgboostDep = "ml.dmlc" % "xgboost4j" % xgboostVersion // scala 2.11 only
+    val xgboostPredictorDep = "biz.k11i" % "xgboost-predictor" % "0.3.1"
```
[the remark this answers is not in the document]
lucagiovagnoli: So the biz.k11i is the original version, you can check how many forks and stars it has compared to the h2oai. I am not sure why h2oai didn't 'github-fork' the project, I think they copied the code and init-ed a new github project, which is very confusing.

So, I've experimented with all of those:
- biz.k11i is the most widely used with 276 stars and 86 forks. Unfortunately it is not maintained anymore and it has a know problem (https://github.com/komiya-atsushi/xgboost-predictor-java/issues/21) because it uses doubles but should use floats (like xgboost4j does). @hollinwilkins was aware of this, so in this PR I'm forcibly casting to floats (like @hollinwilkins  was doing at the time). Check this comment that was left by @hollinwilkins, I dig it back from his PR and re-added it here future reference (https://github.com/combust/mleap/pull/645/files#r416048848)
- h2oai is a 'successor' of biz.k11i which is actively maintained and solves the float vs double bug. I'd love to use it but unfortunately they changed the API in some of the most recent ugprades, so it'd require some work to integrate. I can open an issue in MLEAP to have a look later on, if you'd like.
- On our side we use the Yelp's fork of biz.k11i (https://github.com/Yelp/xgboost-predictor-java) which fixes the double vs float issue and has an additional memory optimization. The problem in adopting the Yelp's biz.k11i version in MLeap is that we don't maintain it as much as h2oai does maintain their 'fork', so maybe their would be more forward-looking? Maybe we can have a look into that in the future..

but personally in the scope of this PR, I think the original biz.k11i is good enough

lucagiovagnoli:
Thanks for reviewing! You raised an excellent question today.
Answers inline:

> @lucagiovagnoli the code changes seem fine. I had minor comments, nothing major.
> 
> I want to check something, we use xgboost-predictor only for classification models, but still use xgboost-4j for regression models, is that right? Why is that?

So, I didn't modify the Regressor in this PR for 2 reasons:
1. keep the change small, not enough time, maybe leave Regressor as a thought exercise :)
2. not sure a Regressor change is really needed.  The Regressor MLeap implementation [here](https://github.com/combust/mleap/blob/eb8ef4e51c2059bd4cbac76db44be58de958840f/mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostRegressionModel.scala#L16) is just a Classifier without a `predictProba()` method. And MLeap rounds that [here](https://github.com/combust/mleap/blob/master/mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/XGBoostClassificationModel.scala#L35) by making `.predict()` enough for regression, I think? If someone really wanted to use the Regressor+Predictor, they could just instantiate a Classifier and call the `.predict()` method. 

> Right now, to use the xgboost-predictor, we'd need to change reference.conf, as you mentioned. I am just thinking if we could instead have two sub-module "mleap-xgboost-4j-runtime" and "mleap-xgboost-predictor" and then depending on the desired implementation, the user would choose between the two? And we can even have a "mleap-xgboost-runtime" base module, that store any code that's common between the two implementations? What do you think?

I think your suggestion makes a lot of sense and would be less confusing than playing with reference.conf - exactly the kind of input I was looking for when asking for your opinion!
I agree that it might be cleaner to separate concerns by creating a separate sbt submodule. We had a small discussion at Yelp about this today, and noticed:

1. On the plus side - as you said - a separate module is easier to use, to test, and it can subclass from shared code, etc.
2. On the down side, a Predictor with its own Op brings some major issues:
- Predictor does *not* implement `store()` primitives, because it's only meant for runtime (see my comment above). So how can we serialize this Op?
- We could  serialize through other Ops (spark-xgboost, runtime-xgboost) but now we incur in loss of flexibility: consumers need to decide which runtime library they want to use.. at serialization time :( That's very restrictive

_Example: say that we train a model using xgboost-spark, the json dump will hold this uniqueName [xgboost.classifier](https://github.com/combust/mleap/blob/master/mleap-xgboost-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostClassificationModelOp.scala#L23). When loading from disk, that will be re-mapped to the xgboost-runtime [here](https://github.com/combust/mleap/blob/master/mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/bundle/ops/XGBoostClassificationOp.scala#L20). A Predictor Op needs  a different uniqueName - say `predictor.classifier` but  how can we map at loading-time which one to use? We're back at using modified reference.conf files.. or we need to decide at serialization time by modifying classes that produce xgboost models._ 

To maintain this spirit of runtime-flexibility, I'm leaning towards choosing via the reference.conf file. Notice that this PR is a no-op for anyone who doesn't care about speed. The change in reference.conf is only needed to switch on to the Predictor, so this weirder user-experience will not reach most MLeap users (xgboost4j  is still the default for xgboost-runtime). 

Let me know what you think!

 
> Could I also please ask you to update the RELEASE_NOTES.md (a first attempt at keeping some release notes) with these changes too? And could you please add any other xgboost related changes that I might have missed? Thank you!

Will do!

lucagiovagnoli:
In the meantime, I added RELEASE_NOTES and a new README for XGBoost-runtime

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/bundle/ops/XGBoostPredictorClassificationOp.scala
ancasarb: makes sense, thank you!

project/Dependencies.scala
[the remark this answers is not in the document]
ancasarb: agreed  biz.k11i is fine, good to have the context.

mleap-xgboost-runtime/src/main/scala/ml/combust/mleap/xgboost/runtime/struct/FVecFactory.scala
ancasarb: 👍

ancasarb:
@lucagiovagnoli What i had in mind: 
1) mleap-xgboost-runtime module - that has any common code between the two implementations
2) mleap-xgboost-4j-runtime module , the reference.conf file is as before

```
ml.combust.mleap.xgboost.ops = [
  "ml.combust.mleap.xgboost.runtime.bundle.ops.XGBoostClassificationOp",
  "ml.combust.mleap.xgboost.runtime.bundle.ops.XGBoostRegressionOp"
]

ml.combust.mleap.registry.default.ops += "ml.combust.mleap.xgboost.ops"
```

XGBoostClassificationOp using the "xgboost.classifier" op name
XGBoostRegressionOp using the "xgboost.regression" op name

3) mleap-xgboost-predictor module with a reference.conf file

```
ml.combust.mleap.xgboost.ops = [
      "ml.combust.mleap.xgboost.runtime.bundle.ops.XGBoostPredictorClassificationOp"   
 ]

ml.combust.mleap.registry.default.ops += "ml.combust.mleap.xgboost.ops"
```

XGBoostPredictorClassificationOp using the "xgboost.classifier" op name as well

We will have the xgboost-spark serialize as before to "xgboost.classifier" and "xgboost.regression" and the choice will come runtime: 
a) if someone wants to use the predictor xgboost classifier, they just import mleap-xgboost-predictor submodule and they're good to go
b) if someone wants to use xgboost-4j, they just import mleap-xgboost-predictor submodule and they're good to go 
c) implementing the regressor later is perfectly fine :) but in the meantime, if someone wants to use the regressor from xgboost-4j and the classifier from xgboost predictor, they would need to import both submodules and to the reference.conf file that you had in mind. but this would be a more rare case, so it would be more straightforward for the general use cases.

what do you think?

lucagiovagnoli:
Oh I see, so if someone does *not* import the xgboost4j, but they import the Predictor, then the big merged reference.conf will only contain the PredictorOp? I haven't tested this but I think that makes a lot of sense. I'll try have a look testing this and splitting the code next week

neptot:
in the source code of XGBoostPredictorClassification
 // Since the Predictor is our performant implementation, we only compute probability for performance reasons.
    val probability = shape.getOutput("probability").map {
      _ => (data: FVec) => Some(model.predictProbabilities(data): Tensor[Double])
    }.getOrElse((_: FVec) => None)

can we use both XGBoostPredictorClassification and XGBoostClassification in the same project, because now we have multi bundles with different op dependency, some depend on XGBoostClassification to support leaf prediction while some not.

neptot:
solved it by register op through code.
use XGBoostClassificationOp as default op without configure through reference.conf
use XGBoostPredictorClassificationOp as op with following code:
  BundleBuilder bundleBuilder = new BundleBuilder();
  ContextBuilder contextBuilder = new ContextBuilder();
  MleapContext mleapContext = contextBuilder.createMleapContext();
  // Register a different Op to change the deserialization class between tests.
  // Use to deserialize with Predictor rather than xgboost4j
  mleapContext.bundleRegistry().register(new XGBoostPredictorClassificationOp());
  Transformer transformer = bundleBuilder.load(modelFile, mleapContext).root();
  //revert to the original Op
  mleapContext.bundleRegistry().register(new XGBoostClassificationOp());

PR #662, opened 2020-04-14
[Issue 617] Upgrade to sklearn 0.22

voganrc:
The `__init__` params for `Imputer` and `OneHotEncoder` have changed in the latest scikit-learn. I updated those classes in MLeap, and all tests pass now. But I still have to check the rest of scikit classes for any changes


python/mleap/sklearn/preprocessing/data.py
```diff
@@ -532,11 +533,19 @@ def __init__(self):
...
 
-        # compile tuples of model attributes to serialize
+        if transformer.strategy == 'most_frequent' or transformer.strategy == 'constant':
+            raise ValueError(f"Imputer strategy `{transformer.strategy}` is not supported by MLeap")
+        if isinstance(transformer.missing_values, str):
+            raise ValueError("Imputer missing values of type `str` are not supported by MLeap")
+        if transformer.add_indicator:
+            raise ValueError("Imputer parameter `add_indicator` is not supported by MLeap")
+        if len(transformer.statistics_.tolist()) != 1:
+            raise ValueError("MLeap Imputer only supports one feature at a time")
```
voganrc, line 543: It seems that on the mleap side you have to specify a single input and output column:
https://github.com/combust/mleap/blob/d276c1c800277a1edcf30dda6e8559c7413b21f8/mleap-runtime/src/test/scala/ml/combust/mleap/runtime/transformer/feature/ImputerSpec.scala#L14

python/mleap/sklearn/preprocessing/data.py
```diff
@@ -532,11 +533,19 @@ def __init__(self):
 
     def serialize_to_bundle(self, transformer, path, model_name):
 
-        # compile tuples of model attributes to serialize
+        if transformer.strategy == 'most_frequent' or transformer.strategy == 'constant':
```
voganrc, line 536: I think we could just set `surrogate_value` if we wanted to implement the `constant_strategy`, because it looks like strategy is ignored during imputation and it just writes whatever constant surrogate value was in the mleap bundle:

https://github.com/combust/mleap/blob/d276c1c800277a1edcf30dda6e8559c7413b21f8/mleap-core/src/main/scala/ml/combust/mleap/core/feature/ImputerModel.scala#L18

But I'm not sure how this would interact with the spark extensions because `constant` is not a valid strategy there:

https://github.com/combust/mleap/blob/fb7493cc662a39eaac5e9bb473f7ea7df71d016a/mleap-spark-extension/src/main/scala/org/apache/spark/ml/mleap/feature/Imputer.scala#L121

python/mleap/sklearn/preprocessing/data.py
```diff
@@ -532,11 +533,19 @@ def __init__(self):
...
+            raise ValueError(f"Imputer strategy `{transformer.strategy}` is not supported by MLeap")
+        if isinstance(transformer.missing_values, str):
+            raise ValueError("Imputer missing values of type `str` are not supported by MLeap")
```
voganrc, line 539: MLeap is all doubles

python/mleap/sklearn/preprocessing/data.py
```diff
@@ -532,11 +533,19 @@ def __init__(self):
...
+        if transformer.add_indicator:
+            raise ValueError("Imputer parameter `add_indicator` is not supported by MLeap")
```
voganrc, line 541: This feature is not on the MLeap side of things... we impute a single column and don't make any new ones

https://scikit-learn.org/stable/modules/generated/sklearn.impute.SimpleImputer.html

python/mleap/sklearn/extensions/data.py
```diff
@@ -51,10 +49,22 @@ def fit_transform(self, X, y=None):
...
         attributes = list()
-        attributes.append(['size', self.n_values_.tolist()[0]])
-        attributes.append(['drop_last', self.drop_last])
+        attributes.append(('category_sizes', self.categories_[0]))
+        if self.handle_unknown == 'error':
+            attributes.append(('handle_invalid', 'error'))
+            attributes.append(('drop_last', self.drop_last))
+        elif self.handle_unknown == 'ignore':
+            attributes.append(('handle_invalid', 'keep'))  # OneHotEncoderModel.scala adds an extra column when keeping invalid data
+            attributes.append(('drop_last', True))         # drop that extra column to match sklearn's ignore behavior
```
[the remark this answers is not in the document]
talalryz: a bit confused here,

In sklearn, when you have drop=None and handle_unknown = "ignore", the category "invalid" will be represented by all zeros. However, when you set drop_last=True, and handle_invalid="keep", the model will drop the last column and add a column for the invalid category.
If the column that was dropped is the same as the handle_invalid column, then the behavior matches sklearn. However, if it isn't then behavior is different. The one-hot encoder in mleap will use all zeros to represent some category other than invalid.
Can you verify that the behavior is correct

python/mleap/sklearn/extensions/data.py
```diff
@@ -109,18 +119,19 @@ def predict(self, X):
 This is because the Imputer both in Spark and MLeap operates on a scalar value and if we were to add a feature extractor in
 front of it, then it would serialize as operating on a tensor and thus, fail at scoring time. 
 """
-class Imputer(SklearnImputer):
+class Imputer(SKLearnImputer):
 
-    def __init__(self, missing_values="NaN", strategy="mean",
-                 axis=0, verbose=0, copy=True, input_features=None, output_features=None):
+    def __init__(self, missing_values=np.nan, strategy="mean",
+                 fill_value=None, verbose=0, copy=True, add_indicator=False,
```
talalryz, line 125: should we have the new kwarg, "add_indicator" be added at the end of the signature so folks using this constructor with positional args don't break

python/mleap/sklearn/base.py
```diff
@@ -58,11 +58,18 @@ def set_input_features(transformer, input_features):
...
-        attributes.append(('intercept', transformer.intercept_.tolist()[0]))
-        attributes.append(('coefficients', transformer.coef_.tolist()[0]))
+        if len(transformer.coef_.shape) == 2 and transformer.coef_.shape[0] == 1:
+            attributes.append(('intercept', transformer.intercept_.tolist()[0]))
+            attributes.append(('coefficients', transformer.coef_.tolist()[0]))
+        elif len(transformer.coef_.shape) == 1:
+            attributes.append(('intercept', transformer.intercept_))
+            attributes.append(('coefficients', transformer.coef_.tolist()))
+        else:
+            raise ValueError("MLeap only supports linear regression on a single target variable")
```
voganrc, line 72: https://github.com/combust/mleap/blob/d276c1c800277a1edcf30dda6e8559c7413b21f8/mleap-core/src/main/scala/ml/combust/mleap/core/regression/LinearRegressionModel.scala#L28

python/mleap/sklearn/base.py
```diff
@@ -58,11 +58,18 @@ def set_input_features(transformer, input_features):
...
 
     def serialize_to_bundle(self, transformer, path, model_name):
+        if transformer.normalize:
+            raise ValueError("MLeap linear regression does not support normalization")
 
-        # compile tuples of model attributes to serialize
         attributes = list()
-        attributes.append(('intercept', transformer.intercept_.tolist()[0]))
-        attributes.append(('coefficients', transformer.coef_.tolist()[0]))
+        if len(transformer.coef_.shape) == 2 and transformer.coef_.shape[0] == 1:
```
voganrc, line 65: `If multiple targets are passed during the fit (y 2D), this is a 2D array of shape (n_targets, n_features), while if only one target is passed, this is a 1D array of length n_features.`
https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html

voganrc: You can fit a linear regressor on a `y` array with shape `(n,1)`, so we still have to consider a 2D `transformer.coef_` array that only denotes one target variable

python/mleap/sklearn/svm.py
```diff
@@ -64,14 +59,14 @@ def set_input_features(transformer, input_features):
         transformer.input_features = input_features
 
     def serialize_to_bundle(self, transformer, path, model_name):
+        if len(transformer.classes_) != 2:
+            raise ValueError("MLeap only supports binary SVM")
```
voganrc, line 63: https://github.com/combust/mleap/blob/d276c1c800277a1edcf30dda6e8559c7413b21f8/mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/classification/SupportVectorMachineOp.scala#L32

python/mleap/sklearn/svm.py
```diff
@@ -38,11 +38,6 @@ def deserialize_from_bundle(self, path, node_name):
     serializer = SimpleSerializer()
     return serializer.deserialize_from_bundle(self, path, node_name)
 
-setattr(SVC, 'op', 'svm')
```
voganrc, line 41: MLeap only supports a linear kernel, so only the `kernel='linear'` option is valid for this class. Because of this, I tried to make the serializer error any time a different `kernel` option was chosen. But even with a linear kernel, the `coef_` matrix of this class is a read-only derived attribute:

https://stackoverflow.com/questions/34827838/scikit-learn-numpy-changing-the-value-of-an-read-only-variable

So we would have to reconstruct the coefficients from `model.dual_coef_` and `model.support_vectors_` when deserializing. Rather than doing this, I thought we should just remove `SVC` support all-together, as the `kernel='linear'` case is already covered by `LinearSVC` anyway

voganrc: Also... there were no tests for this class. And if you try to actually use MLeap to serialize `SVC` it fails . So it's safe to say nobody is using this feature

PR #664, opened 2020-04-22
[Issue 617] Upgrade to scikit-learn version 0.22

voganrc:
The interfaces for `OneHotEncoder` and `Imputer` have changed in the latest scikit-learn


lucagiovagnoli (dismissed):
1. Thanks for adding so many tests, they all look great. Some of the tests that you moved could have almost lived in their own PR. Make sure tests pass on github, I noticed people tend not to review if the build doesn't pass
2. It'd be useful to know which of the changes in the Imputer and OneHotEncoder are due to the sklearn 0.22 ugprade and which ones are just raising better exceptions than we should have been raising before. Basically the PR is both improving current code and upgrading the APi and that makes it a bit harder to read. One way to tackle this problem in the future could be to separate commits and good description of what each commit brings in in the "Conversation" Github field. Such us:
- commit 1: improvements on older code - BASELINE
- commit 2: Move code to 0.22
- commit 3: add all tests

mleap-runtime/src/e2e/resources/sklearn_scripts/one_hot_encoder.py
```diff
@@ -0,0 +1,85 @@
...
+
+    to_df = NDArrayToDataFrame(['first_0', 'first_1', 'first_2', 'last_0', 'last_1', 'last_2'])
+
+    union_pipeline = Pipeline([(feature_union.name, feature_union), (to_df.name, to_df)])
+    union_pipeline.fit(df)
+    union_pipeline.mlinit()
+
+    union_pipeline.serialize_to_bundle(args.bundle_dir, "one_hot_encoder_bundle", init=True)
+    df.to_csv(args.original_csv_path, index=False)
+    union_pipeline.transform(df).to_csv(args.transformed_csv_path, index=False)
```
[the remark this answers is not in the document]
voganrc: This script is not itself a test but generates the bundle file and csv files used in end to end testing. You can see it gets invoked in SKLearnOneHotEncoderParitySpec.scala

python/mleap/sklearn/preprocessing/data.py
```diff
@@ -624,15 +629,32 @@ def __init__(self):
...
-        attributes.append(('drop_last', False))
+        attributes.append(('size', singular_categories.size))
+        if transformer.handle_unknown == 'error':
+            attributes.append(('handle_invalid', 'error'))
+            attributes.append(('drop_last', transformer.drop_last))
+        elif transformer.handle_unknown == 'ignore':
+            attributes.append(('handle_invalid', 'keep'))  # OneHotEncoderModel.scala adds an extra column when keeping invalid data
+            attributes.append(('drop_last', True))         # drop that extra column to match sklearn's ignore behavior
+        else:
+            raise ValueError(f"Unrecognized `handle_unknown` value {self.handle_unknown}")
```
talalryz, line 656: I think that we can move this to a helper function

python/mleap/sklearn/preprocessing/data.py
```diff
@@ -648,20 +670,19 @@ def serialize_to_bundle(self, transformer, path, model_name):
     def deserialize_from_bundle(self, transformer, node_path, node_name):
 
         attributes_map = {
-            'size': 'n_values_'
+            'size': 'categories_',
```
talalryz, line 673: We should make a note that models serialized with older versions of mleap will not be deserialized correctly

python/tests/sklearn/preprocessing/test_one_hot_encoder.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def tearDown(self):
+        shutil.rmtree(self.tmp_dir)
+
+    def test_one_hot_encoder_serialization_fails_on_multiple_feature_columns(self):
+        self.oh_data = np.hstack((self.oh_data, self.oh_data))  # make two feature columns
+
+        ohe = OneHotEncoder(handle_unknown='error')
+        ohe.mlinit(prior_tf=self.le, output_features='{}_one_hot_encoded'.format(self.le.output_features))
+        ohe.fit(self.oh_data)
```
talalryz, line 49: I'm a bit surprised that this doesn't throw an error here since handle_unknown="error"

python/tests/sklearn/preprocessing/test_one_hot_encoder.py
```diff
@@ -0,0 +1,145 @@
...
+        ohe.serialize_to_bundle(self.tmp_dir, ohe.name)
+
+        node_name = "{}.node".format(ohe.name)
+        ohe_ds = OneHotEncoder()
+        ohe_ds.deserialize_from_bundle(self.tmp_dir, node_name)
+
+        self.oh_data[2][0] = 3  # Add an unknown category
+
+        expected = ohe.transform(self.oh_data).todense()
+        actual = ohe_ds.transform(self.oh_data)
```
talalryz, line 143: Why do we need to call toDense() on ohe.transform but not on ohe_ds.transform, I would expect the model behavior to be the same pre and post serialization.

python/tests/sklearn/preprocessing/test_one_hot_encoder.py
```diff
@@ -0,0 +1,145 @@
...
+        ohe.mlinit(prior_tf=self.le, output_features='{}_one_hot_encoded'.format(self.le.output_features))
+        ohe.fit(self.oh_data)
+
+        ohe.serialize_to_bundle(self.tmp_dir, ohe.name)
+
+        node_name = "{}.node".format(ohe.name)
+        ohe_ds = OneHotEncoder()
+        ohe_ds.deserialize_from_bundle(self.tmp_dir, node_name)
+
+        self.oh_data[2][0] = 3  # Add an unknown category
```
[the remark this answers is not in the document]
voganrc: The problem here is that I need `oh_data` to form an invalid range. Like this:
`[0, 1, 2, 4]`
But if I were to pass strings through the `LabelEncoder`, it will always map them to a valid category range. Like this:
`[0, 1, 2, 3]`
So this test is meant for the case where somebody loads a csv and tries to one-hot encode an integer feature. Say number of bedrooms in a house. No house has zero bedrooms, so we would force them to convert the data to category indices before one-hot encoding

python/tests/sklearn/preprocessing/test_one_hot_encoder.py
voganrc: handle_unknown only matters if you call `transform` on labels that weren't seen during `fit`. So this doesn't raise here because `fit` is what determines which classes are unknown in the first place

python/tests/sklearn/preprocessing/test_one_hot_encoder.py
voganrc: Good question! The MLeap deserialization code for OneHotEncoder has always set `sparse=False` when loading a bundle. 

https://github.com/combust/mleap/blob/d703c8dd5e7b93c7f6c8b45fa6c9aadb9c026a92/python/mleap/sklearn/preprocessing/data.py#L658

I believe it's to make it easy to convert into a dataframe in this function:

https://github.com/combust/mleap/blob/d703c8dd5e7b93c7f6c8b45fa6c9aadb9c026a92/python/mleap/sklearn/preprocessing/data.py#L778

Because otherwise you would have to call `from_spmatrix`:

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sparse.from_spmatrix.html

But I'm not 100% sure this is the reason why

Either way, the default for the scikit-learn OneHotEncoder is to set `sparse=True`. So it seems that this normally results in an implicit conversion. Another option would be to error unless users explicitly set `sparse=False` every time they instantiate a OneHotEncoder, but that would be pretty annoying. I'm not sure if you have an opinion on which is better...

python/mleap/sklearn/preprocessing/data.py
voganrc: Where would be the best place to do this? In the release notes?

ancasarb:
Thank you @voganrc for the PR, just a small note that I will take a look at this this week. Thank you!

python/mleap/sklearn/preprocessing/data.py
```diff
@@ -624,18 +629,29 @@ def __init__(self):
...
-        attributes.append(('drop_last', True))
+        attributes.append(('size', singular_categories.size))
         if transformer.handle_unknown == 'ignore':
+            # MLeap's OneHotEncoderModel adds an extra column when keeping invalid data
+            # so we drop that extra column to match sklearn's ignore behavior
             attributes.append(('handle_invalid', 'keep'))
+            attributes.append(('drop_last', True))
         else:
             attributes.append(('handle_invalid', 'error'))
+            attributes.append(('drop_last', False))
```
voganrc, line 652: @ancasarb When going through this code again, I realized that we can't always do `attributes.append(('drop_last', True))` like in your PR. If we do, then `transformer.handle_unknown == 'error'` will cause MLeap to drop the last feature column, but scikit will still have that column

python/mleap/sklearn/preprocessing/data.py
```diff
@@ -651,20 +667,19 @@ def serialize_to_bundle(self, transformer, path, model_name):
...
+        if transformer.handle_unknown == 'keep':
+            transformer.handle_unknown = 'ignore'
+        transformer.drop_idx_ = None
 
-        # Set Feature Indices
-        n_values = np.hstack([[0], [transformer.n_values_]])
-        indices = np.cumsum(n_values)
-        transformer.feature_indices_ = indices
-        transformer.active_features_ = range(0, transformer.n_values_)
+        transformer.sparse = False
```
voganrc, line 681: This is potentially problematic because the default value for the `sparse` parameter in `OneHotEncoder.__init__` is `True`. So we would be silently changing this during deserialization. I don't know if this is better or worse than erroring every time users don't explicitly pass `sparse=False`, because that would be really annoying

python/tests/sklearn/preprocessing/data_test.py
```diff
@@ -369,68 +368,6 @@ def label_encoder_deserializer_test(self):
         self.assertEqual(le.input_features, label_encoder_tf.input_features)
         self.assertEqual(le.output_features, label_encoder_tf.output_features[0])
 
-    def one_hot_encoder_serializer_test(self):
```
voganrc, line 372: These tests were moved to `python/tests/sklearn/preprocessing/one_hot_encoder_test.py`

python/tests/sklearn/preprocessing/data_test.py
```diff
@@ -486,60 +423,6 @@ def feature_extractor_test(self):
         self.assertEqual(expected_model['attributes']['input_shapes']['data_shape'][1]['isNullable'],
                      model['attributes']['input_shapes']['data_shape'][1]['isNullable'])
 
-    def imputer_test(self):
```
voganrc, line 489: This test was moved to `python/tests/sklearn/preprocessing/one_hot_encoder_test.py`

python/mleap/sklearn/extensions/data.py
```diff
@@ -109,18 +90,20 @@ def predict(self, X):
 This is because the Imputer both in Spark and MLeap operates on a scalar value and if we were to add a feature extractor in
 front of it, then it would serialize as operating on a tensor and thus, fail at scoring time. 
 """
-class Imputer(SklearnImputer):
+class Imputer(SKLearnImputer):
 
-    def __init__(self, missing_values="NaN", strategy="mean",
-                 axis=0, verbose=0, copy=True, input_features=None, output_features=None):
+    def __init__(self, input_features, output_features,
```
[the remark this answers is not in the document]
voganrc: @ancasarb What do you think about this order of arguments?

python/requirements.txt
```diff
@@ -2,6 +2,6 @@ numpy>=1.8.2
 six>=1.10.0
 scipy>=0.13.0b1
 pandas>=0.18.1, <= 0.24.2
-scikit-learn>0.18.dev0,<0.20.0
+scikit-learn>=0.21.0,<0.23.0
```
voganrc, line 5: `StandardScaler` breaks in version 0.23 😢

voganrc:
@lucagiovagnoli @ancasarb Bumping this

ancasarb:
taking a look @voganrc

mikaylaedwards:
Hi all, any updates on when this will be merged? Scikit is broken for me when installed from master but works when I installed from the above PR. Also, mleap is awesome! Thank you!

python/mleap/bundle/serialize.py
```diff
@@ -88,7 +88,7 @@ def get_mleap_model(self, transformer, attributes_to_serialize=None):
 
             elif isinstance(value, np.ndarray):
                 attributes[name] = {
-                    "double": list(value.flatten()),
+                    "double": value.flatten().tolist(),
```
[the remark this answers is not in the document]
ancasarb: 👍

python/mleap/sklearn/extensions/data.py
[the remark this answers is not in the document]
ancasarb: this order seems fine.

python/mleap/sklearn/extensions/data.py
```diff
@@ -61,18 +63,20 @@ def predict(self, X):
...
+                 missing_values=np.nan, strategy="mean",
+                 fill_value=None, verbose=0, copy=True, add_indicator=False):
+        self.op = "imputer"
```
ancasarb, line 71: i am a bit confused as to how did this work before, was there some place else where this was being done?

python/mleap/sklearn/preprocessing/data.py
ancasarb: release notes has a breaking change section that we could use for this. i've added them in for the future release, 0.17.0 in this PR https://github.com/combust/mleap/pull/736.

python/mleap/sklearn/preprocessing/data.py
ancasarb: yeah, not sure about this. i think for now, maybe we could leave it as is setting `sparse=False` and see if it becomes an issue.

python/mleap/sklearn/preprocessing/data.py
ancasarb: 👍 thanks for looking into this again, it's been quite a bit of a pain to get this right.

ancasarb (commented):
changes look good to me, i had one small question only. 

could you please also fix the conflicts on the release notes? sorry about this, not sure how to better do this at the moment. 

and we could get this PR merged afterwards.

thank you!

python/mleap/sklearn/extensions/data.py
voganrc: Oh I don't think this is necessary actually because we call `setattr` on the parent `SKLearnImputer`:
https://github.com/combust/mleap/pull/664/files?file-filters%5B%5D=.py&file-filters%5B%5D=.txt#diff-47b26af574b6dfeab149aeaa450acefa3ed9e37daa90407b90a8076ded3a5db4R121

voganrc:
@ancasarb I took another look at this PR and responded to your review feedback.
I think it should be ready to go now.

RELEASE_NOTES.md
```diff
@@ -5,8 +5,10 @@
 ### New Features
 
 ### Bug Fixes
+- Fix OneHotEncoder Python serialization
```
ancasarb, line 8: could i please ask you to move this to the breaking changes section, or mention it there as well? 

i think Talal mentioned that we should make a note that models serialized with older versions of mleap will not be deserialized correctly on upgrade, so i think we should mention that.

ancasarb (approved):

RELEASE_NOTES.md
voganrc: Hi Anca, I don't think this is actually a breaking change...

The line that Talal was referring to was this one here:
https://github.com/combust/mleap/pull/664/files#diff-47b26af574b6dfeab149aeaa450acefa3ed9e37daa90407b90a8076ded3a5db4L654

Old bundles would still have the `{"size": n}` attribute serialized on disk. We are only changing what this gets deserialized into because the `n_values_` property has been deprecated.

ancasarb: 👍

PR #575, opened 2019-10-11
Build against spark3.0.0-preview

WeichenXu123:
Build against spark3.0.0-preview

Fix some code which is broken under spark 3.0:
* update some breeze operator
* Update IDFModel initialization
* Update Tree node initialization

Disable some spec test which is broken, these may be due to model/transformer behavior change, fix them require much work, we could fix them in separate PR. 
* HashingTermFrequencyParitySpec
* GaussianMixtureParitySpec
* MaxAbsScalerParitySpec
* MinMaxScalerParitySpec

PR #665, opened 2020-04-24
Spark 3.0 / Scala 2.12 compatible build

WeichenXu123:
Take over https://github.com/combust/mleap/pull/661

* Add numerical tolerance in comparing float/double in test.
* Restore some code change in ScalerModel

Test against spark-3.0.0-SNAPSHOT


ancasarb:
@WeichenXu123 is this the latest PR with the spark 3.0 related changes? does this mean https://github.com/combust/mleap/pull/661 and https://github.com/combust/mleap/pull/575 could be potentially closed? just wanted to double check.

WeichenXu123:
@ancasarb Yes. This PR takes over #661 and #575

mleap-spark-testkit/src/main/scala/org/apache/spark/ml/parity/SparkParityBase.scala
```diff
@@ -18,6 +19,10 @@ import ml.combust.mleap.spark.SparkSupport._
 import ml.combust.mleap.runtime.transformer.Pipeline
 import resource._
 
+import org.apache.spark.sql.Row
+import org.apache.spark.ml.linalg.{Vector, Vectors}
+import org.apache.spark.ml.util.TestingUtils._
```
mengxr, line 24: explicit import `relTol`

mleap-spark-testkit/src/main/scala/org/apache/spark/ml/parity/SparkParityBase.scala
```diff
@@ -131,10 +136,38 @@ abstract class SparkParityBase extends FunSpec with BeforeAndAfterAll {
     }
   }
 
+  def checkRowWithRelTol(actualAnswer: Row, expectedAnswer: Row, relTol: Double) = {
```
mengxr, line 139: do not duplicate the `relTol` name

mleap-spark-testkit/src/main/scala/org/apache/spark/ml/parity/SparkParityBase.scala
```diff
@@ -131,10 +136,38 @@ abstract class SparkParityBase extends FunSpec with BeforeAndAfterAll {
...
+      s"actual answer length ${actualAnswer.length} != " +
+        s"expected answer length ${expectedAnswer.length}")
+
+    // TODO: support struct types?
+    actualAnswer.toSeq.zip(expectedAnswer.toSeq).foreach {
+      case (actual: Double, expected: Double) =>
+        assert(actual ~== expected relTol relTol)
+      case (actual: Vector, expected: Vector) =>
+        assert(actual ~= expected relTol relTol)
+      case (actual: Array[Double], expected: Array[Double]) =>
```
mengxr, line 150: Array -> Seq?

mleap-spark-testkit/src/main/scala/org/apache/spark/ml/parity/SparkParityBase.scala
```diff
@@ -131,10 +136,38 @@ abstract class SparkParityBase extends FunSpec with BeforeAndAfterAll {
...
-    val mleapElems = mleapDataset.collect()
-    assert(sparkElems sameElements mleapElems)
+    val sparkCols = sparkDataset.columns.toSeq
+    assert(mleapDataset.columns.toSet === sparkCols.toSet)
+    val sparkRows = sparkDataset.collect()
+    val mleapRows = mleapDataset.select(sparkCols.map(col): _*).collect()
+    var i = 0
+    for ((sparkRow, mleapRow) <- sparkRows.zip(mleapRows)) {
+      checkRowWithRelTol(sparkRow, mleapRow, relTol)
+      i += 1
```
mengxr, line 169: `i` not used

mleap-spark-testkit/src/main/scala/org/apache/spark/ml/parity/SparkParityBase.scala
WeichenXu123: Cannot only import `relTol` there're other internal implicits classes need import.
The `TestingUtils` class only contains functionality about equality test with numerical error. So usually we should import `TestingUtils._`

mleap-spark-testkit/src/main/scala/org/apache/spark/ml/parity/SparkParityBase.scala
WeichenXu123: Seq don't work.
Will match Seq[String] to case Seq[Double], I don't kown why...

WeichenXu123:
@mengxr Ready. Remaining following issues we can address them in follow-up PRs:
* NaiveBayes model support Gauss type. We need update code in `ml.combust.mleap.core.classification.NaiveBayesModel` to make it also support Gauss type.

* `ml.feature.HashingTF` change hash function in spark 3.0. We should make it backwards compatible in mleap. But current `ml.feature.HashingTF` implementation make `hashFunc` to be a private var, mleap has no way to set it when loading old version HashTF model. We need create a spark PR to address it.

* `org.apache.spark.ml.bundle.tree.decision.SparkNodeWrapper` only support "gini" impurity ("gini" is hardcoded), we should update code to support all kinds of impurity.

* Release a xgboost version for scala_2.12, so that mleap can support it with spark 3.0 + scala_2.12.

.travis.yml
```diff
@@ -18,25 +18,30 @@ jobs:
...
       install:
         - pip install "virtualenv<20.0.0" tox  # Bug https://github.com/pypa/virtualenv/issues/1551
       script:
-        - sbt +compile
-        - tox -c python/tox.ini -e py36 -v
+        - sbt -DsparkVersion=2.4.5 -DscalaVersion=2.11.12 compile
+        - tox -c python/tox-spark2.4.ini -e py37 -v
 
-    - name: "Python 3.7 tests"
+    - name: "Python 3.7 + scala 2.12 + spark 3.0 tests"
```
lucagiovagnoli, line 44: Rather than having this complex matrix running tests to support all of these combinations, I wonder if it makes sense to just upgrade as a clean cut. we could just create an additional point release  moving to spark3.0  (say mleap 0.20) and whoever wants spark2.4 just uses mleap 0.16

but I'd refer to @hollinwilkins  and @ancasarb for more insights

WeichenXu123: We hope mleap is backwards compatible, e.g., user train and saved model by old version mleap, we can load old version model via new version mleap and run model prediction.

.travis.yml
```diff
@@ -18,25 +18,30 @@ jobs:
...
       script:
-        - sbt +compile
-        - tox -c python/tox.ini -e py27 -v
+        - sbt -DsparkVersion=2.4.5 -DscalaVersion=2.11.12 compile
+        - tox -c python/tox-spark2.4.ini -e py27 -v
 
-    - name: "Python 3.6 tests"
-      language: python
-      python: 3.6
+    - name: "Python 3.7 + scala 2.11 + spark 2.4 tests"
```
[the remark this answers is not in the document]
WeichenXu123: If py3.6 needed, I can add it back.

.travis.yml
WeichenXu123: @lucagiovagnoli If we make two branch mleap, one for spark 2.4.x and another for spark 3.x, then mleap new features and bugfix added in future should go into the two branches code, it make things more troublesome. So that I think best approach will still be make mleap master code support both spark 2.4.x and spark 3.x, in this PR I make a thin wrapper to address spark ML API incompatibility between spark 2.4.x and spark 3.x

mleap-core/src/test/scala/ml/combust/mleap/core/feature/MaxAbsScalerModelSpec.scala
```diff
@@ -12,10 +14,9 @@ class MaxAbsScalerModelSpec extends FunSpec {
     val scaler = MaxAbsScalerModel(Vectors.dense(Array(20.0, 10.0, 10.0, 20.0)))
 
     it("Scales the vector based on absolute max value"){
-      val inputArray = Array(15.0, -5.0, 5.0, 19.0)
-      val expectedVector = Array(0.75, -0.5, 0.5, 0.95)
-
-      assert(scaler(Vectors.dense(inputArray)).toArray.sameElements(expectedVector))
+      val inputVector = Vectors.dense(15.0, -5.0, 5.0, 19.0)
+      val expectedVector = Vectors.dense(0.75, -0.5, 0.5, 0.95)
+      assert(scaler(inputVector) ~= expectedVector relTol 1E-3)
```
mengxr, line 19: Is 1e-3 too loose?

mleap-spark/build.sbt
```diff
@@ -2,3 +2,11 @@ import ml.combust.mleap.{Dependencies, Common}
 
 Common.defaultMleapSettings
 Dependencies.spark
+
+unmanagedSourceDirectories in Compile += {
+  if (Dependencies.sparkVersion.startsWith("3.0.")) {
+    baseDirectory.value / "src" / "spark-shims" / "spark-3.0"
```
mengxr, line 8: Maybe out of scope here: how are we going to make the release?

mengxr:
@WeichenXu123 I think none of the issues you mentioned at https://github.com/combust/mleap/pull/665#issuecomment-621258623 should be addressed in this PR. Could you please open one issue for each?

mengxr:
@ancasarb @WeichenXu123 @lucagiovagnoli I think the main question to answer is that what we need to support in the next MLeap release. To me, we should definitely support the following:

Spark 3.0, Scala 2.12, Python 3. Backward compatibility: MLeap can load previously saved MLeap bundles and preserve previous behavior, e.g., hashing functions.

It is unclear to me whether we still need to support Spark 2.4 or Scala 2.11. Users can use previous release of MLeap to convert Spark ML models into MLeap bundles, right?

The issue with supporting both is that we have to introduce shims like in this PR because there are breaking changes in Spark 3.0. Then how do we make a new release? The jar for Spark 2.4 would be different for the jar for Spark 3.0. It adds lot of complications. Your thoughts?

WeichenXu123:
@mengxr 
> Users can use previous release of MLeap to convert Spark ML models into MLeap bundles, right?

If this, we can remove the spark 2.4/3.0 shims and make a release for spark 2.4 first and then make a release for spark 3.0.

If users need to smoothly transition from spark 2.4 to 3.0, and hope in spark 2.4 also benefit from mleap new features, then spark 2.4/3.0 shims is required.

I am neutral on this and it depends on what user needed..

mengxr:
I like the idea to insert a Spark 2.4 release with other new features before a Spark 3.0 release. But creating versions with "_spark_2.4" and "_spark_3.0" suffixes is very error-prone because dependency solver like maven cannot tell that they are not compatible.

WeichenXu123:
@mengxr I create a PR which remove spark shims and depend on spark 2.4.5
https://github.com/combust/mleap/pull/669
We can release a version mleap-1.6 compatible spark 2.4.5 and then release a new version mleap compatible with spark 3.0

WeichenXu123:
@mengxr Ready for review.
Note: Xgboost still has issue (mainly kyro lib incompatible issue). We need build against this PR version xgboost:
https://github.com/dmlc/xgboost/pull/5648

Test command:
```
# clone this PR branch xgboost first: https://github.com/dmlc/xgboost/pull/5648
cd /path/to/xgboost
git submodule init
git submodule update
cd jvm-packages/
mvn clean install -DskipTests # will locally install xgboost version 1.0.0-spark3

# clone this PR branch mleap
cd /path/to/mleap
git submodule init
git submodule update
sbt "test" "mleap-executor-tests/test" "mleap-benchmark/test" "mleap-xgboost-runtime/test" "mleap-xgboost-spark/test"
```

@ancasarb Could you create a branch for spark 3.0 ? Then I can submit this PR against that branch.

WeichenXu123:
@mengxr 
Ready.
Some updates:
* fixed the kryo dependency (exclude "com.esotericsoftware.kryo" % "kryo")
* merge master
* revert commit https://github.com/combust/mleap/commit/06a84b62f4cbe9227af07ed78cfc649f7c317309 in master. (This break xgboost unit test when build with spark 3.0, @lucagiovagnoli @ancasarb Could you also help fix it for spark 3.0 after spark-3.0 branch created ?)

ancasarb:
branch created here https://github.com/combust/mleap/tree/spark-3.0.0. please let me know if it's ok.

WeichenXu123:
@mengxr @ancasarb Re-submit PR here https://github.com/combust/mleap/pull/678

PR #666, opened 2020-04-24
Math Unary pyspark wrapper

lucagiovagnoli:
**Main change:**
- Added a MathUnary pyspark wrapper (ref. https://github.com/combust/mleap/issues/570)
- Removed py27 tests. This means we're not testing against py27 anymore. py27 is deprecated, so I don't think we should hold on to it.

**Minor changes**
- Compile spark-extension jars for python tests _once before each python test stage_. This allows for removing the slower `+sbt/compile`.
- Pass the `SCALA_CLASS_PATH` as an environment variable down to tox and python tests removing all `os.path.join('..',.'..')`. This is more robust to diverse environments and test execution paths and allowed to simplify a lot of code.
- Use the python `tempfile` package to create temporary dirs during testing. Not all systems have permissions to manually write to '/tmp/'

@ancasarb for review -- I'm also going to tackle other pyspark wrappers from https://github.com/combust/mleap/issues/570  as these are pretty urgent for us.
@talalryz @voganrc


python/mleap/pyspark/__init__.py
```diff
@@ -11,3 +14,4 @@
 sys.modules['pyspark.ml'].mleap = mleap
 sys.modules['pyspark.ml'].mleap.feature = sys.modules['mleap.pyspark.feature']
 sys.modules['pyspark.ml'].mleap.feature.StringMap = StringMap
+sys.modules['pyspark.ml'].mleap.feature.MathUnary = MathUnary
```
lucagiovagnoli, line 17: Small note, I'm still looking at this - if there's a better way to do it.

python/mleap/pyspark/feature/math_unary.py
```diff
@@ -0,0 +1,84 @@
...
+
+            scalaMathUnaryModel = _jvm().ml.combust.mleap.core.feature.MathUnaryModel(scalaUnaryOperation)
+
+            self._java_obj = self._new_java_obj(
+                "org.apache.spark.ml.mleap.feature.MathUnary",
+                self.uid,
+                scalaMathUnaryModel,
+            )
+
+        self._setDefault()
```
ancasarb, line 63: for my knowledge only, what does this do?

ancasarb, line 63: for my knowledge only, what does this do?

python/mleap/pyspark/feature/string_map.py
```diff
@@ -53,7 +92,17 @@ def from_dataframe(cls, labels_df, inputCol, outputCol, handleInvalid='error', d
         See StringMap() for other params.
         """
         assert isinstance(labels_df, DataFrame), 'labels must be a DataFrame, got: {}'.format(type(labels_df))
-        labels_dict = {row[0]: float(row[1]) for row in
-                       labels_df.select([inputCol, outputCol]).collect()}
-        return cls(labels_dict, inputCol=inputCol, outputCol=outputCol, handleInvalid=handleInvalid,
-                   defaultValue=defaultValue)
+        labels_dict = {
+            row[0]: float(row[1])
+            for row in
```
ancasarb, line 97: nitpick, is the extra space required?

python/mleap/pyspark/feature/string_map.py
```diff
@@ -38,13 +43,47 @@ def validate_args():
...
+        )
+
+        self._setDefault()
+        self.setParams(inputCol=inputCol, outputCol=outputCol)
+
+    @keyword_only
+    def setParams(self, inputCol=None, outputCol=None):
+        """
+        setParams(self, inputCol=None, outputCol=None)
+        Sets params for this MathUnary.
```
ancasarb, line 71: Copy paste :)

python/tests/pyspark/string_map_test.py
```diff
@@ -19,85 +22,113 @@
...
+            handleInvalid='keep',
+            defaultValue=-1.0
+        ).transform(self.input)
+        expected = StringMapTest.spark.createDataFrame([['a', 'b', -1.0]], OUTPUT_SCHEMA)
         assert_df(expected, result)
 
     def test_map_missing_value_error(self):
         with self.assertRaises(Py4JJavaError) as error:
-            StringMap({'z': 1.0}, 'key_col', 'value_col').transform(self.input).collect()
+            StringMap(
```
ancasarb, line 68: nitpick, is the extra space required?

python/tests/pyspark/string_map_test.py
```diff
@@ -19,85 +22,113 @@
...
+                outputCol='value_col'
+            ).transform(self.input).collect()
         self.assertIn('java.util.NoSuchElementException: Missing label: a', str(error.exception))
 
     def test_map_from_dataframe(self):
-        labels_df = self.spark.createDataFrame([['a', 1.0]], 'key_col: string, value_col: double')
-        result = StringMap.from_dataframe(labels_df, 'key_col', 'value_col').transform(self.input)
-        expected = self.spark.createDataFrame([['a', 'b', 1.0]], OUTPUT_SCHEMA)
+        labels_df = StringMapTest.spark.createDataFrame([['a', 1.0]], 'key_col: string, value_col: double')
+        result = StringMap.from_dataframe(
```
ancasarb, line 77: same here

travis/scala_classpath_for_python.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/bash
+
+# Compile classes used in python tests
+sbt mleap-spark-extension/compile
+sbt mleap-spark/compile
+
+# Export the complete dependencyClasspath into an env. variable
+export SCALA_CLASS_PATH=\
+"$(sbt --error 'set showSuccess := false' mleap-spark-extension/printClassPath 2> /dev/null)"
```
juhoautio, line 9: Now if this needs to be run manually any way, there's no need for a custom sbt task at all.

This single command provided by sbt should do it:

`sbt "export mleap-spark-extension/runtime:fullClasspath"`

This is nice because this includes the paths to `classes` folders so you can edit scala code, let your IDE compile it, and rerun python tests without visiting sbt/command line.

lucagiovagnoli: Oh, interesting, that's useful, I'll update that, thanks!

.travis.yml
```diff
@@ -46,9 +38,11 @@ jobs:
       dist: xenial  # Travis won't install py37 on trusty
       install:
         - pip install "virtualenv<20.0.0" tox  # Bug https://github.com/pypa/virtualenv/issues/1551
+      before_script:
+        - source travis/scala_classpath_for_python.sh
       script:
-        - sbt +compile
-        - tox -c python/tox.ini -e py37 -v
+        - echo $SCALA_CLASS_PATH
+        - SCALA_CLASS_PATH=$SCALA_CLASS_PATH tox -c python/tox.ini -e py37 -v
```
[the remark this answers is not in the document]
lucagiovagnoli: Sure, I'll document the steps in a comment in the next revision. 

Summing up, I just run this once:
`source travis/scala_classpath_for_python.sh`

It loads the classpath string into SCALA_CLASS_PATH which is then passed through to Python so that the classpath is available to the sparkContext for all test runs. 

For tests you can run:
`tox -c python/tox.ini -e py37 -v`
or also you can run any test singularly, the environemnt variable is always available.

(If you check the tox.ini file under `passenv`, I'm forwarding SCALA_CLASS_PATH there.)

python/tests/pyspark/lib/spark_session.py
```diff
@@ -13,34 +13,9 @@ def spark_session():
...
-    Read classpath for mleap-spark-extension with the locally compiled classes.
-
-    Note that pyspark 2.4.4 comes with spark jars with specific scala version, for example:
-
-        mleap/python/venv/lib/python3.7/site-packages/pyspark/jars/spark-core_2.11-2.4.4.jar
-
-    Thus, pyspark is incompatible with scala 2.12 and only works with scala 2.11.
-
-    This means that mleap pyspark wrappers have not been tested against scala 2.12. However, they may still work as long
-    as all jars are with the same scala version.
```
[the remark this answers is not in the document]
lucagiovagnoli: That makes sense. I think I'm only running 2.11 at the moment. I'll re-introduce that comment in the next revision!

.travis.yml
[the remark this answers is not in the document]
juhoautio: Thanks.

Also I’m wondering specifically if there’s some handy way to make the env variable available in an IDE like PyCharm. Especially if the classpath changes because of something, would I need to copy the new env var to my IDE manually? Previously everything was bound to the sbt compilation so it was naturally always up to date. I understand the reasons why you’re changing this, but still, I hope there would be some clean way to setup IDE without too much manual maintenance.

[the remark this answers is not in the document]
lucagiovagnoli: So there's a few ways I would do this in PyCharm. 
1. set an env. variable in vanilla PyCharm, explanation here: https://stackoverflow.com/a/42708480/2544874
2. You can also use this JetBrains plugin: https://plugins.jetbrains.com/plugin/7861-envfile
it basically lets you load environment variables from a file. When I use that plugin I ensure I generate an hidden .env_variables.env file as it's best to avoid committing  workflow related files and procedures into the project.
3. Otherwise you can run from command line (this is my preferred method cause it lets me test things cross-platform, including what happens on Travis) :
```
source travis/scala_classpath_for_python.sh (this only once in the terminal)
tox -c python/tox.ini -e py37 -v
```

I can't think of a circumstance where the classpath would change "often". If it does, you can `source` again, or you could have your own script to regenerate the .env file in option 2) or reset option 1)

python/mleap/pyspark/feature/math_unary.py
lucagiovagnoli: So, the entire structure of this `JavaTransformer` is taken from the apache/spark project. Check their huge file of pyspark `features` here (Normalizer, NGram etc.):
https://github.com/apache/spark/blob/4a64901ab7176ad3d41c4598dea1e73cd92e969e/python/pyspark/ml/feature.py#L1827

As you can see I'm inheriting from JavaTransformer and implementing methods, calling `_setDefault()` like they are doing (I think it's a no-op but if the mother-project is calling it, I wanted to be consistent).

Too bad Spark doesn't provide MathUnary pyspark wrappers already.. This is also the reason for the [controversial](https://github.com/combust/mleap/pull/666/files#r414778794):
```python
sys.modules['pyspark.ml.mleap'] = mleap
```

Unfortunately [this code](https://github.com/apache/spark/blob/3ce1dff7ba4799295a2bf89e6df35e900afc587b/python/pyspark/ml/wrapper.py#L244) in Apache/spark
 searches for the python classes under the `pyspark` module. They don't let us inject the location of our wrapper's module from the outside, so we need to monkeypatch the code to make the MathUnary appear as if it was in the apache/spark project. It's similar to what MLeap does with the reference.conf file but less flexible 

One solution to this could be to subclass the JavaParams I linked above and override the entire `_from_java()` method, but I'm not sure how everyone feels about that..

python/mleap/pyspark/__init__.py
lucagiovagnoli: One solution to this could be to subclass the JavaParams from [here](https://github.com/apache/spark/blob/3ce1dff7ba4799295a2bf89e6df35e900afc587b/python/pyspark/ml/wrapper.py#L244) and override the entire _from_java() method, but I'm not sure how everyone feels about that..

python/mleap/pyspark/feature/string_map.py
lucagiovagnoli: I think the space is only in the github UI, there's no newline between line 97 and 98, or is that what you meant ?

python/tests/pyspark/string_map_test.py
lucagiovagnoli: What extra space? If you mean putting all args on separate lines, I'm following python linters.
It'd be nice to introduce linter checks in `make test` for the python subproject, I'll try get to it while doing the `MathBinary`

python/mleap/pyspark/feature/string_map.py
```diff
@@ -53,7 +92,17 @@ def from_dataframe(cls, labels_df, inputCol, outputCol, handleInvalid='error', d
...
-                   defaultValue=defaultValue)
+        labels_dict = {
+            row[0]: float(row[1])
+            for row in
+           labels_df.select([inputCol, outputCol]).collect()
+        }
+
+        return cls(
+            labels=labels_dict,
+	    inputCol=inputCol,
```
talalryz, line 103: looks misaligned

python/mleap/pyspark/feature/string_map.py
lucagiovagnoli: ops

lucagiovagnoli:
> Had a couple of questions/comments, mostly nitpicks.
> 
> Changes look good to me. Could you please add the changes to release notes as well? Thank you!

Done, I've also added documentation in the `README.md`, thanks everyone for reviewing!

travis/scala_classpath_for_python.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/bash
+
+# Compile classes used in python tests
+sbt mleap-spark-extension/compile
+sbt mleap-spark/compile
+
+# Export the complete dependencyClasspath into an env. variable
+export SCALA_CLASS_PATH=\
+"$(sbt --error 'export mleap-spark-extension/runtime:fullClasspath' 2> /dev/null)"
```
juhoautio, line 9: I was looking for some docs in README on how to run python tests locally. Would that make sense, or are developers supposed to figure it out from this file?

.travis.yml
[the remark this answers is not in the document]
juhoautio: Thanks for those tips – the env file plugin can make this convenient enough at least.

Do you think some of those tips could make their way into a README?

> I can't think of a circumstance where the classpath would change "often"

Agreed. The point is though, I'd like it to be convenient when it happens. If I pull changes from repo that includes some change to the classpath, I then run sbt compile which is standard, but for python there's a unofficial step to remember / look up.

If nothing else, should the README include this that code snippet:

```
source travis/scala_classpath_for_python.sh (this only once in the terminal)
tox -c python/tox.ini -e py37 -v
```

?

travis/scala_classpath_for_python.sh
lucagiovagnoli: Generally, I'd figure out how to run tests from the project setup (follow the steps in a Makefile, or do what tox.ini does). In MLeap we're expected to replicate locally how Travis runs tests, so I'd open the [.travis.yaml](https://github.com/combust/mleap/pull/666/files#diff-354f30a63fb0907d4ad57269548329e3R27) file and read:
```yaml
        - source travis/scala_classpath_for_python.sh
        - SCALA_CLASS_PATH=$SCALA_CLASS_PATH tox -c python/tox.ini -e py36 -v
```
Also the python line `os.env['SCALA_CLASS_PATH']` should trigger a `find` search for `SCALA_CLASS_PATH` in the project.

I generally prefer holding the code accountable for documentation purposes because the code can't go out of date. A README is good until it's up to date.

However, if you feel it could be useful, I'll add a section. I'm only afraid the `.travis.yaml` file could soon change (specially after https://github.com/combust/mleap/pull/665) rendering the README useless. In the meantime, I just noticed I didn't add this in the python Makefile, so I'll add a section there which might also help in documenting this

.travis.yml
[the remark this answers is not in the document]
lucagiovagnoli: That's reasonable!

I think Makefile have this job generally, MLeap doesn't have a top-level Makefile but uses Travis, so I hoped that reading Travis should be enough, but it's probably very helpful to add all of this in the python specific Makefile too as many devs refer to Makefiles as a first thing.

If you feel strongly about a README, I'll add it in there too

travis/scala_classpath_for_python.sh
juhoautio: > I generally prefer holding the code accountable for documentation purposes because the code can't go out of date. A README is good until it's up to date.

That's very true. What you described makes sense (open `.travis.yaml` etc.) 👍 

I think for a quick start I would still prefer simple instructions in README. It could be something like this:

----

To compile & run tests:

```
sbt compile test (not sure about this command though?)
```

To run python tests:

- NOTE: `sbt compile` must have been run before this

```
cd python
make (requires some arg?)
```

But if it seems it can't be kept up to date, I won't insist :) It should be ok to find this by checking the travis file.

In any case this shouldn't be a concern for this PR's scope. Thanks for taking the time to explain this to me.

Makefile
```diff
@@ -0,0 +1,9 @@
+SHELL := /bin/bash
+
+.PHONY: py36_test py37_test
+
+py36_test:
+	source travis/scala_classpath_for_python.sh && make -C python py36_test
+
+py37_test:
+	source travis/scala_classpath_for_python.sh && make -C python py37_test
```
juhoautio, line 9: I suppose the nice separation of concerns would be that `.travis.yaml` & `travis/` (directory) only contain travis-specific conf. Right?

In that sense it seems wrong that `Makefile` uses `scala_classpath_for_python.sh` from under `travis/` when on the other hand `.travis.yaml` uses the `Makefile`. It's not exactly a circular dependency, but mentally yes. Should `scala_classpath_for_python.sh` be moved out of `travis/`?

lucagiovagnoli: Oh yes, totally. That's a remnant from when the call was in `.travis.yaml`, I'll move it out

travis/scala_classpath_for_python.sh
lucagiovagnoli: So, right now Travis runs Scala tests using this long sequence of commands:
```
  sbt "+ test" "+ mleap-executor-tests/test" "+ mleap-benchmark/test" "mleap-xgboost-runtime/test" "mleap-xgboost-spark/test" "+ publishSigned" "mleap-xgboost-runtime/publishSigned" "mleap-xgboost-spark/publishSigned" "mleap-serving/docker:publish" "mleap-spring-boot/docker:publish"
```
(taken from https://github.com/combust/mleap/blob/master/travis/travis.sh). That might change in #665. I don't think that's in the README.

After my last revision, python tests are as simple as running:
```
make py37_test
```
from the top-level folder. Sourcing is only required if running make from the inner folder

python/Makefile
```diff
@@ -1,3 +1,11 @@
+SHELL := /bin/bash
+
+ifndef SCALA_CLASS_PATH
+$(error SCALA_CLASS_PATH for python tests is not set. Please check out \
+the top-level Makefile on how to source scala_classpath_for_python.sh)
```
juhoautio, line 5: Nice 👍 

For the hasty (like me), `spark_session.py` could also have an assertion with a helpful message like this?

python/Makefile
```diff
@@ -1,3 +1,11 @@
+SHELL := /bin/bash
+
+ifndef SCALA_CLASS_PATH
+$(error SCALA_CLASS_PATH for python tests is not set. Please check out \
```
juhoautio, line 4: IMHO `SCALA_CLASS_PATH` is a bit too generic name. I would rather use a more specific name, for example `MLEAP_RUNTIME_CLASSPATH`.

python/README.md
```diff
@@ -83,3 +99,17 @@ You can also take your scikit pipelines and deploy them to your Spark cluster, b
...
 * [Scikit-Learn](http://mleap-docs.combust.ml/getting-started/scikit-learn.html)
+
+## Contributions
+Contributions are welcome! Make sure all python tests pass.
+You can run them from the top-level makefile:
+```bash
+make py37_test
+```
+
+If you'd rather use the inner `python/Makfile`, remember to source SCALA_CLASS_PATH by running:
```
juhoautio, line 110: typo: Makfile

PR #522, opened 2019-04-30
 fix deserialization in spark

tammosminia:
Improved tests to look inside a pipeline for more models to also compare those.
Fixed all the resulting failed tests.
- Some are actual errors where part of the read model is ignored
- don't compare parameters that are only of importance during training

PR #674, opened 2020-05-06
Serialize handleInvalid param for Bucketizer

WeichenXu123:
Serialize handleInvalid param for Bucketizer

### DIFF
--- a/mleap-core/src/main/scala/ml/combust/mleap/core/feature/BucketizerModel.scala
+++ b/mleap-core/src/main/scala/ml/combust/mleap/core/feature/BucketizerModel.scala
@@ -11,13 +11,24 @@ import ml.combust.mleap.core.types.{ScalarType, StructType}
   * @param splits splits used to determine bucket
   */
 @SparkCode(uri = "https://github.com/apache/spark/blob/v2.0.0/mllib/src/main/scala/org/apache/spark/ml/feature/Bucketizer.scala")
-case class BucketizerModel(splits: Array[Double]) extends Model {
+case class BucketizerModel(splits: Array[Double], handleInvalid: String = "error") extends Model {
   def apply(feature: Double): Double = {
-    binarySearchForBuckets(splits, feature)
+    if (handleInvalid == "skip") {
+      -1
+    } else {
+      binarySearchForBuckets(splits, feature, keepInvalid = handleInvalid == "keep")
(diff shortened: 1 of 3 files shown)


lucagiovagnoli:
I found lots of these Transformers with similar issues, where some parameter was not serialized. If we had more tests, we could ensure e2e correctness and find out more of these. I've started writing some Parity tests like these ones: https://github.com/combust/mleap/blob/502cbef25692c6362f1b785ea98ed5cbe0d05880/mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModelParitySpec.scala

It'd be great if you could add a few small ones for the code you added

ancasarb:
@lucagiovagnoli We actually have parity tests for the params equality, in the SparkParityBase

```
    def checkParamsEquality(original: Transformer, deserialized: Transformer): Unit = {
      val ignoredParams = unserializedParams
      assert(original.params.length == deserialized.params.length)
      original.params.zip(deserialized.params).foreach {
        case (param1, param2) => if(!ignoredParams.contains(param1.name)) {
          assert(original.isDefined(param1) == deserialized.isDefined(param2),
            s"spark transformer param ${param1.name} is defined ${original.isDefined(param1)} deserialized is ${deserialized.isDefined(param2)}")

          if (original.isDefined(param1)) {
            val v1Value = original.getOrDefault(param1)
            val v2Value = deserialized.getOrDefault(param1)

            v1Value match {
              case v1Value: Array[_] => assert(v1Value sameElements v2Value.asInstanceOf[Array[_]])
              case _ => assert(v1Value == v2Value, s"$param1 is not equivalent")
            }
          }
        }
      }
    }
```

A couple releases ago, we went through all transformers and try to fix the serialization to/from Spark, in this PR https://github.com/combust/mleap/pull/522. There may be some small issues still when both spark/mleap had a default for some of the parameters, though I've tried to go through the *ParitySpec.scala, to add a transformer that didn't use defaults for most part. Just something to be aware of :) 

In this case, "handleInvalid" for Bucketizer is added in this updated version of Spark, 2.4.5.

lucagiovagnoli:
Oh, I didn't see the checkParamsEquality - so the test  caught it when upgrading to 2.4.5? perfect!

Issue #667, opened 2020-04-28
OneHotEncoder Python Extension is Broken


voganrc:
In scikit-learn's OneHotEncoder, unknown labels are represented as a row of all zeros:

https://github.com/scikit-learn/scikit-learn/blob/4755ae76d2df10bbf41bc93fb7083b0142ef1044/sklearn/preprocessing/_encoders.py#L220

Whereas in MLeap, unknown labels get their own column appended onto the one-hot encoded matrix:

https://github.com/combust/mleap/blob/3ee879fddb9d134c16ced719e2b0a3668f3071e2/mleap-core/src/main/scala/ml/combust/mleap/core/feature/OneHotEncoderModel.scala#L26

This means that in order to replicate the behavior of the scikit-learn OneHotEncoder, we have to drop this extra column. The way to do this in MLeap is to include `drop_last=True` in the bundle serialization.

But because the point of the OneHotEncoder Python extension is to allow users to choose the `drop_last` param:

https://github.com/combust/mleap/blob/d703c8dd5e7b93c7f6c8b45fa6c9aadb9c026a92/python/mleap/sklearn/extensions/data.py#L57

That class is useless. The serialized `drop_last` needs to always be `True`, because we always have to remove the invalid column from the MLeap implementation. User's don't actually have a choice whether they want to drop the last feature column.

lucagiovagnoli:
Thanks for opening this issue. We can tackle this small sub-problem here rather than in #664 as that one is already huge

ancasarb:
@voganrc @lucagiovagnoli Yes, the recent changes in OneHotEncoder/OneHotEncoderEstimator have been a bit of a nightmare in terms of maintaining backwards compatibility. Could you please take a look at https://github.com/combust/mleap/pull/677, which removes this and attempts to fix scikit-learn onehotencoder by serializing dropLast=true for parity? Thank you!

PR #673, opened 2020-05-06
Serialize handleInvalid param of VectorIndexer into mleap model

WeichenXu123:
Serialize handleInvalid param of VectorIndexer into mleap model

PR #677, opened 2020-05-10
Changes from parity testing

ancasarb:
These changes (the ones around OneHotEncoder) are after I've tested roughly 50 models serialized using MLeap 0.10.3 for parity (especially around OneHotEncoder). Attempts to fix https://github.com/combust/mleap/issues/667 as well. 

Main changes:
- make all transformers that support handleInvalid use the trait HandleInvalid (including the ones that only support only keep and error)
- ensure transformers like Bucketizer(s) support skip correctly, by skipping the row in the result
- attempt again to fix parity of old/deprecated in 2.4.x OneHotEncoder between Spark, Mleap and Scikit-learn
- documentation updates (removing also outdated xgboost documentation)
- updating the release notes

Note: to be merged after https://github.com/combust/mleap/pull/673 is reviewed and merged, so that VectorIndexer can use HandleInvalid as well.

### DIFF
--- a/mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/OneHotEncoderOp.scala
+++ b/mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/OneHotEncoderOp.scala
@@ -33,7 +33,8 @@ class OneHotEncoderOp extends MleapOp[OneHotEncoder, OneHotEncoderModel] {
       if (model.getValue("size").nonEmpty) {
         // Old version of 1HE.
         OneHotEncoderModel(categorySizes = Array(model.value("size").getLong.toInt),
-          dropLast = model.value("drop_last").getBoolean)
+          dropLast = model.value("drop_last").getBoolean,
+          handleInvalid = HandleInvalid.default)
       } else {
         // New version of 1HE.
         OneHotEncoderModel(
--- a/mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/OneHotEncoderOp.scala
+++ b/mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/OneHotEncoderOp.scala
@@ -39,7 +39,7 @@ class OneHotEncoderOp extends MleapOp[OneHotEncoder, OneHotEncoderModel] {
         // New version of 1HE.
         OneHotEncoderModel(
           categorySizes = model.value("category_sizes").getIntList.toArray,
-          handleInvalid = HandleInvalid.fromString(model.value("handle_invalid").getString),
+          handleInvalid = HandleInvalid.fromString(model.value("handle_invalid").getString, false),
           dropLast = model.value("drop_last").getBoolean)
       }
     }
--- a/mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/OneHotEncoderOp.scala
+++ b/mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/OneHotEncoderOp.scala
@@ -33,8 +33,6 @@ class OneHotEncoderOp extends MleapOp[OneHotEncoder, OneHotEncoderModel] {
       if (model.getValue("size").nonEmpty) {
         // Old version of 1HE.
         OneHotEncoderModel(categorySizes = Array(model.value("size").getLong.toInt),
-          // assume HandleInvalid.Keep always since old version of 1HE didn't have a way to specify this
-          handleInvalid = HandleInvalid.Keep,
           dropLast = model.value("drop_last").getBoolean)
       } else {
         // New version of 1HE.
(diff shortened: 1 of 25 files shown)


ancasarb:
@WeichenXu123 @lucagiovagnoli I will create a spark 3.0.0 branch, so that @WeichenXu123 can start raising PRs and merging changes into that branch.

I will do this first thing tomorrow, after merging these three changes:
- https://github.com/combust/mleap/pull/673
- https://github.com/combust/mleap/pull/677
- https://github.com/combust/mleap/pull/645

Does this sound like a good plan? Please let me know if you have any worries or concerns. Thank you both for all your help!

mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/OneHotEncoderOp.scala
```diff
@@ -30,18 +30,19 @@ class OneHotEncoderOp extends MleapOp[OneHotEncoder, OneHotEncoderModel] {
 
     override def load(model: Model)
                      (implicit context: BundleContext[MleapContext]): OneHotEncoderModel = {
+      val dropLast = model.value("drop_last").getBoolean
       if (model.getValue("size").nonEmpty) {
         // Old version of 1HE.
+        val handleInvalid = if (dropLast) HandleInvalid.Keep else HandleInvalid.default
```
voganrc, line 36: You hard-code `attributes.append(('drop_last', True))` in `data.py`, so this removes the choice between `HandleInvalid.Keep` and `HandleInvalid.Error`.

Instead, we should just read whatever the bundle has inside of it, and have the Python class serialize:
```
if transformer.handle_unknown == 'ignore':
    attributes.append(('handle_invalid', 'keep'))
else:
    attributes.append(('handle_invalid', 'error'))
attributes.append(('drop_last', True))
```

mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/OneHotEncoderOp.scala
```diff
@@ -30,18 +30,19 @@ class OneHotEncoderOp extends MleapOp[OneHotEncoder, OneHotEncoderModel] {
...
         OneHotEncoderModel(categorySizes = Array(model.value("size").getLong.toInt),
-          // assume HandleInvalid.Keep always since old version of 1HE didn't have a way to specify this
-          handleInvalid = HandleInvalid.Keep,
-          dropLast = model.value("drop_last").getBoolean)
+          dropLast = dropLast,
+          handleInvalid = handleInvalid)
```
voganrc, line 39: nit: order of arguments is different from the `else` branch. Also you create a `handleInvalid` variable here but inline it in the assignment below

ancasarb: Fixed.

mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/OneHotEncoderOp.scala
ancasarb: You're right, I've amended the code.

ancasarb:
@voganrc @lucagiovagnoli thank you for your review, I've addressed the comments, could you please take another look? thanks again!

mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/OneHotEncoderOp.scala
```diff
@@ -30,18 +30,22 @@ class OneHotEncoderOp extends MleapOp[OneHotEncoder, OneHotEncoderModel] {
...
-        OneHotEncoderModel(categorySizes = Array(model.value("size").getLong.toInt),
-          // assume HandleInvalid.Keep always since old version of 1HE didn't have a way to specify this
-          handleInvalid = HandleInvalid.Keep,
-          dropLast = model.value("drop_last").getBoolean)
+        OneHotEncoderModel(
+          categorySizes = Array(model.value("size").getLong.toInt),
+          handleInvalid =
+            model.getValue("handle_invalid").map(_.getString).map(HandleInvalid.fromString(_, false))
+              //this maintains backwards compatibility for models that don't have handle_invalid serialized
+              .getOrElse(if (dropLast) HandleInvalid.Keep else HandleInvalid.default),
```
voganrc, line 41: Should we make this `.getOrElse(HandleInvalid.Keep)` to match the hard-coded `handleInvalid = HandleInvalid.Keep` from before? 

Let's say there's an old bundle that had `drop_last = False` in it. If they were to run that bundle through the old code they would get an extra column for invalid data. But with the new code the same bundle would error when it sees invalid data.

ancasarb: Unfortunately, HandleInvalid.error needs to be the default for old spark models to be backwards compatible. I made previously the change to just use HandleInvalid.Keep thinking it works for Spark models too, but it doesn't :( It's been a bit of a nightmare to get this 1he updates working.

PR #679, opened 2020-05-13
Interaction pyspark wrapper in MLeap 2.4.x

lucagiovagnoli:
**NOTE: I'm putting this out here, in case anyone needs it, I don't expect it to be necessarily merged due to the nature of this special monkeypatching.**

------------------
**Main change:**
* I'm copying the pyspark3.0 [Interaction](https://github.com/apache/spark/blob/branch-3.0/python/pyspark/ml/feature.py#L1702) to MLeap. Since "Interaction" is only available in spark **3.0**, this PR allows anyone to use it in MLeap **2.4.x**

**Special remarks:**
* Unfortunately, I had to monkeypatch `sys.module` in order for this to work, because [this line](https://github.com/apache/spark/blob/master/python/pyspark/ml/wrapper.py#L244) in pyspark does a `replace` for a specific string. More details in why-comments in the code.
* The monkeypatching masks the "Interaction" import from pyspark3.0, so I added a try-catch to explicitly fail whenever MLeap upgrades to spark3.0 to remember removing the file . 
* Note that this monkeypatching is different from the others (MathUnary etc) because those add the `mleap` namespace to pyspark, and therefore don't mask any pyspark objects.

### DIFF
--- a/python/mleap/pyspark/__init__.py
+++ b/python/mleap/pyspark/__init__.py
@@ -8,6 +8,9 @@ from mleap.pyspark.feature.string_map import StringMap
 from mleap.pyspark.feature.math_binary import MathBinary
 from mleap.pyspark.feature.math_unary import MathUnary
 
+# These monkeypatches make it so mleap.feature appears to be in pyspark.ml
+# This should be safe unless pyspark uses the namespace
+# pyspark.ml.mleap (unlikely). Then we would mask those modules (undesirable).
 sys.modules['pyspark.ml.mleap'] = mleap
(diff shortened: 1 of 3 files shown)


ancasarb:
@talalryz @voganrc should we close this for now and re-add this once we've upgraded to Spark 3.0? what do you think?

talalryz:
> @talalryz @voganrc should we close this for now and re-add this once we've upgraded to Spark 3.0? what do you think?

makes sense to me!

ancasarb:
closing for now, as agreed.

PR #678, opened 2020-05-11
Spark 3.0 / Scala 2.12 compatible build

WeichenXu123:
Spark 3.0 / Scala 2.12 compatible build.

Take over https://github.com/combust/mleap/pull/661

* Add numerical tolerance in comparing float/double in test.
* Restore some code change in ScalerModel
* fixed the kryo dependency (exclude "com.esotericsoftware.kryo" % "kryo")
* merge master
* revert commit 

Test against spark-3.0.0-rc1

PR #680, opened 2020-05-13
add Spark test dependencies and change parity tooling to allow numerical tolerance

WeichenXu123:
add Spark test dependencies and change parity tooling to allow numerical tolerance.

### DIFF
--- a/mleap-core/src/test/scala/ml/combust/mleap/core/feature/MaxAbsScalerModelSpec.scala
+++ b/mleap-core/src/test/scala/ml/combust/mleap/core/feature/MaxAbsScalerModelSpec.scala
@@ -16,7 +16,7 @@ class MaxAbsScalerModelSpec extends FunSpec {
     it("Scales the vector based on absolute max value"){
       val inputVector = Vectors.dense(15.0, -5.0, 5.0, 19.0)
       val expectedVector = Vectors.dense(0.75, -0.5, 0.5, 0.95)
-      assert(scaler(inputVector) ~= expectedVector relTol 1E-3)
+      assert(scaler(inputVector) ~= expectedVector relTol 1E-9)
     }
 
     it("Has the right input schema") {
--- a/mleap-core/src/test/scala/ml/combust/mleap/core/feature/MaxAbsScalerModelSpec.scala
+++ b/mleap-core/src/test/scala/ml/combust/mleap/core/feature/MaxAbsScalerModelSpec.scala
@@ -4,6 +4,8 @@ import ml.combust.mleap.core.types.{StructField, TensorType}
 import org.apache.spark.ml.linalg.Vectors
 import org.scalatest.FunSpec
 
+import org.apache.spark.ml.util.TestingUtils._
+
 /**
   * Created by mikhail on 9/18/16.
   */
@@ -12,10 +14,9 @@ class MaxAbsScalerModelSpec extends FunSpec {
     val scaler = MaxAbsScalerModel(Vectors.dense(Array(20.0, 10.0, 10.0, 20.0)))
 
     it("Scales the vector based on absolute max value"){
-      val inputArray = Array(15.0, -5.0, 5.0, 19.0)
-      val expectedVector = Array(0.75, -0.5, 0.5, 0.95)
-
-      assert(scaler(Vectors.dense(inputArray)).toArray.sameElements(expectedVector))
+      val inputVector = Vectors.dense(15.0, -5.0, 5.0, 19.0)
+      val expectedVector = Vectors.dense(0.75, -0.5, 0.5, 0.95)
+      assert(scaler(inputVector) ~= expectedVector relTol 1E-3)
     }
 
     it("Has the right input schema") {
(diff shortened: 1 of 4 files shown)


WeichenXu123:
@lucagiovagnoli @ancasarb 
According to comments here https://github.com/combust/mleap/pull/678#issuecomment-627422465
Could you review & merge this PR to master first, and then merge mleap master to mleap branch:spark-3.0.0 ?

Thanks!!

project/Dependencies.scala
```diff
@@ -82,11 +82,12 @@ object Dependencies {
     val akkaStreamTestKit = "com.typesafe.akka" %% "akka-stream-testkit" % akkaVersion % "test"
     val junit = "junit" % "junit" % "4.12" % "test"
     val junitInterface = "com.novocode" % "junit-interface" % "0.10" % "test"
-    val spark = Compile.spark.map(_ % "test")
+    val spark = Compile.spark.map(_ % "test") ++ Compile.spark.map(_ % "test" classifier "tests")
   }
 
   object Provided {
     val spark = Compile.spark.map(_.excludeAll(ExclusionRule(organization = "org.scalatest"))).map(_ % "provided")
+    val sparkTestLib = spark.map(_ classifier "tests")
```
mengxr, line 90: We don't need test jars in runtime. So they shouldn't under "Provided".

project/Dependencies.scala
```diff
@@ -82,11 +82,12 @@ object Dependencies {
     val akkaStreamTestKit = "com.typesafe.akka" %% "akka-stream-testkit" % akkaVersion % "test"
     val junit = "junit" % "junit" % "4.12" % "test"
     val junitInterface = "com.novocode" % "junit-interface" % "0.10" % "test"
-    val spark = Compile.spark.map(_ % "test")
+    val spark = Compile.spark.map(_ % "test") ++ Compile.spark.map(_ % "test" classifier "tests")
```
mengxr, line 85: I would list test jars under a different name, e..g, "sparkTest".

project/Dependencies.scala
```diff
@@ -101,25 +102,25 @@ object Dependencies {
...
 
-  val core = l ++= Seq(sparkMllibLocal, jTransform, Test.scalaTest)
+  val core = l ++= Seq(sparkMllibLocal, jTransform, Test.scalaTest) ++ Test.spark
 
   def runtime(scalaVersion: SettingKey[String]) = l ++= (Seq(Test.scalaTest, Test.junit, Test.junitInterface, commonsIo) ++ scalaReflect.modules(scalaVersion.value))
 
   val sparkBase = l ++= Provided.spark ++ Seq(Test.scalaTest)
 
-  val sparkTestkit = l ++= Provided.spark ++ Seq(scalaTest)
+  val sparkTestkit = l ++= Provided.spark ++ Provided.sparkTestLib ++ Seq(scalaTest)
```
mengxr, line 111: You should get the dependencies from Test instead of "Provided".

mleap-spark-testkit/src/main/scala/org/apache/spark/ml/parity/SparkParityBase.scala
```diff
@@ -131,10 +136,45 @@ abstract class SparkParityBase extends FunSpec with BeforeAndAfterAll {
...
+            assert(actualElem == expectedElem, s"$actualElem did not equal $expectedElem")
+        }
+      case (actual: Row, expected: Row) =>
+        checkRowWithRelTol(actual, expected, eps)
+      case (actual, expected) =>
+        assert(actual == expected, s"$actual did not equal $expected")
+    }
+  }
+
+  var relTolEps: Double = 1E-3
```
ancasarb, line 169: I agree that comparing floats/doubles for exact equality could lead to issues but I think the order of magnitude here is too big, perhaps something like 1E-9 or 1E-12?

mleap-core/src/test/scala/ml/combust/mleap/core/feature/MaxAbsScalerModelSpec.scala
```diff
@@ -12,10 +14,9 @@ class MaxAbsScalerModelSpec extends FunSpec {
     val scaler = MaxAbsScalerModel(Vectors.dense(Array(20.0, 10.0, 10.0, 20.0)))
 
     it("Scales the vector based on absolute max value"){
-      val inputArray = Array(15.0, -5.0, 5.0, 19.0)
-      val expectedVector = Array(0.75, -0.5, 0.5, 0.95)
-
-      assert(scaler(Vectors.dense(inputArray)).toArray.sameElements(expectedVector))
+      val inputVector = Vectors.dense(15.0, -5.0, 5.0, 19.0)
+      val expectedVector = Vectors.dense(0.75, -0.5, 0.5, 0.95)
+      assert(scaler(inputVector) ~= expectedVector relTol 1E-3)
```
ancasarb, line 19: See my comment below, 1E-3 seems too big, is there an actual reason why this cannot be something like 1E-9 or 1E-12?

project/Dependencies.scala
WeichenXu123: Oh, this is a special case, referenced by sub-project mleap-spark-testkit (this project has no test code, the main src code need reference spark testing API)

project/Dependencies.scala
WeichenXu123: No. explained here https://github.com/combust/mleap/pull/680#discussion_r425543010

WeichenXu123:
@mengxr @ancasarb Ready.

mengxr:
Merged into master and spark-3.0.0.

Issue #659, opened 2020-04-08
Include a __version__ in python mleap module


Ben-Epstein:
For governance purposes and following https://www.python.org/dev/peps/pep-0396/

can you include a __version__ for your module? It would make things much easier to track, thanks!

ancasarb:
We have this here https://github.com/combust/mleap/blob/master/python/mleap/version.py, does this need updating or something?

Ben-Epstein:
@ancasarb Sorry for the late reply. This is why I've brought up the issue.
![Screen Shot 2020-05-19 at 9 42 34 AM](https://user-images.githubusercontent.com/22605641/82333737-1cd3e300-99b5-11ea-81b3-80546e1654e0.png)

I see the version in the code, but something about its format isn't correct as I cannot access it.

ancasarb:
Yeah, at the moment, it seems you need to do 
```
from mleap.version import __version__ as mleap_version
print(mleap_version)
```

I'll see if we can improve this.

ancasarb:
Merged PR to master, will include in next release. Closing this.

Issue #441, opened 2018-11-02
Unable to load persisted PipelineModel containing mleap Imputer: Column imputeValue must be old Vector type to be converted to new type but got DoubleType


bleskowsky:
We've been experiencing this issue when trying to load a saved PipelineModel that contains the mleap Imputer implementation as one of its stages. I have a simplified program here that replicates the issue on my local environment. This is using **Spark 2.3.0** and **mleap 0.12.0**.

Sample program:
```scala
import org.apache.spark.ml.mleap.feature.Imputer
import org.apache.spark.ml.{Pipeline, PipelineModel}
import org.apache.spark.sql.SparkSession

case class Datapoint(id: Int, score: Option[Double])

object MLeapExample {
  def main(args: Array[String]): Unit = {

    val spark = SparkSession
      .builder()
      .appName("MLeap Export")
      .master("local")
      .getOrCreate()

    import spark.implicits._

    // some sample training data (score is nullable)
    val p1 = Datapoint(1, Some(1.0))
    val p2 = Datapoint(2, null)
    val p3 = Datapoint(3, Some(9.0))

    val training = List(p1, p2, p3).toDF

    training.show()

    val imputer = new Imputer()
      .setInputCol("score")
      .setOutputCol("imputed_score")
      .setStrategy("median")

    val pipeline = new Pipeline()
      .setStages(Array(imputer))

    val model = pipeline.fit(training)

    model.write.overwrite().save("trained_model")

    val loaded_model = PipelineModel.load("trained_model")

  }
}
```
The input training data looks as follows:
```
+---+-----+
| id|score|
+---+-----+
|  1|  1.0|
|  2| null|
|  3|  9.0|
+---+-----+
```
This is a very tiny sample of a column in our data that we would typically attempt to impute a value for (score with strategy median). The type of the column to impute is either a double or null. 

Running this example via IntelliJ and spark-submit i.e.

`spark-submit --class MLeapExample --packages ml.combust.mleap:mleap-spark-base_2.11:0.12.0,ml.combust.mleap:mleap-spark_2.11:0.12.0,ml.combust.mleap:mleap-spark-extension_2.11:0.12.0 mleap-export_2.11-0.1.jar`

leads to the following stacktrace:
```
Exception in thread "main" java.lang.IllegalArgumentException: requirement failed: Column imputeValue must be old Vector type to be converted to new type but got DoubleType.
        at scala.Predef$.require(Predef.scala:224)
[...]

Issue #420, opened 2018-09-18
Unable to save serialized model to HDFS


AashimaYuthika:
I have written a basic script with pyspark for writing a serialized model to HDFS. Here's the code:
```import mleap.pyspark
from mleap.pyspark.spark_support import SimpleSparkSerializer
from pyspark import SparkContext
from pyspark.sql import SparkSession
from pyspark.ml.feature import VectorAssembler, StandardScaler, OneHotEncoder, StringIndexer
from pyspark.ml import Pipeline, PipelineModel
from pyspark.sql import Row

sc = SparkContext(appName='CTRPrediction')
spark = SparkSession(sc).builder.getOrCreate()

l = [('Alice', 1), ('Bob', 2)]
rdd = sc.parallelize(l)
Person = Row('name', 'age')
person = rdd.map(lambda r: Person(*r))
df2 = spark.createDataFrame(person)
df2.collect()

string_indexer = StringIndexer(inputCol='name', outputCol='name_string_index')

feature_assembler = VectorAssembler(inputCols=[string_indexer.getOutputCol()], outputCol="features")

feature_pipeline = [string_indexer, feature_assembler]

featurePipeline = Pipeline(stages=feature_pipeline)

fittedPipeline = featurePipeline.fit(df2)

fittedPipeline.serializeToBundle("jar:hdfs:///tmp/pyspark.example.zip", fittedPipeline.transform(df2))
```
Running it on a Yarn Cluster like this
`spark-submit --master yarn --deploy-mode cluster --conf spark.yarn.appMasterEnv.PYTHON_EGG_CACHE="/tmp/.cache/" --py-files mleap-0.11.0.dev0-py2.7.egg --jars mleap-spark-base_2.11-0.11.0.jar,mleap-spark_2.11-0.11.0.jar --packages ml.combust.mleap:mleap-spark_2.11:0.11.0 aashima/mleap_test.py`

This is the error that I am getting.
```Traceback (most recent call last):
  File "mleap_test.py", line 34, in <module>
    fittedPipeline.serializeToBundle("jar:hdfs:///tmp/pyspark.example.zip", fittedPipeline.transform(df2))
  File "/mnt/yarn/usercache/hadoop/appcache/application_1524545386558_0665/container_1524545386558_0665_02_000001/mleap-0.11.0.dev0-py2.7.egg/mleap/pyspark/spark_support.py", line 25, in serializeToBundle
[...]

PR #681, opened 2020-05-17
Small fixes + implement MLWritable/MLReadable for math transformers as well

ancasarb:
Fixes:
- #420
- #441
- #659


lucagiovagnoli:
The code change looks good.

I only have one question, [here](https://github.com/combust/mleap/blob/5c40bb6094dbcca95b0bb3dc8f37d5bc218e80f7/python/tests/pyspark/feature/math_binary_test.py#L114) I am already able to serialize and deserialize MathUnary and MathBinary. Why do we need MLReadable and MLWriteable? Is that to implement Spark-like 'native' serialization? And when should someone use this rather than that ?

ancasarb:
@lucagiovagnoli Yes, MLReadable and MLWriteable is needed to implement Spark-like 'native' serialization, I've only really seen the need come up for some of our existing notebooks for e.g. in Databricks that were already doing that, so it's just a not too complicated way not to break that by implementing those two interfaces.

python/mleap/__init__.py
```diff
@@ -0,0 +1,3 @@
+from mleap.version import version
+
+__version__ = version
```
ancasarb, line 3: @lucagiovagnoli let me know if you know a better way to do this.

lucagiovagnoli: I think the convention is to have this `__init__.py` file contain:
```
__version__ = 0.16.0-SNAPSHOT
```
rather than importing it from another version.py file.

So just move the contents of version.py in this file :)

ancasarb: I tried that, but because it's used in setup.py, it didn't work to put it here, because i think when setup.py runs, this doesn't exist yet.

lucagiovagnoli:
Was this the last PR needed before release ?

ancasarb:
yes, just released 0.16.0 :)

PR #692, opened 2020-06-04
Update shading rules for databricks runtime assembly

mengxr:
Shade more dependencies.


WeichenXu123:
Looks good. I am trying to write code to run test cases against the fat jar.

PR #698, opened 2020-06-16
Change spring boot docker port

hessam-shm:
Currently, in the documentation it is said to run docker file with -p 8080:8080 while in application.yml file the port is 65327. This should change to be in accord with the documentation.


ancasarb:
hey @hessam-shm, we've gone back and fourth on this, but i think what we settled on was that the default grpc port is 65328 and can be overridden via MLEAP_GRPC_PORT and the default http port is 65327 and can be overridden via MLEAP_HTTP_PORT.

thanks for pointing out the issue with the documentation, i've updated this to match now, thank you again!

PR #701, opened 2020-06-23
U/talal/fix math binary pyspark

talalryz:
This PR makes fixes the following:

1. The pyspark wrapper for math binary did not handle defaults correctly. You can see it in the PR that introduces the pyspark wrapper for mathbinary: https://github.com/combust/mleap/pull/672/ . The PR fixes this issue and brings the pyspark wrapper on par with base spark. Tests have been added to ensure correctness.

2. Fixes the following serialization direction: python serialize -> scala deserialize. This is scala deserialization reading defaults as 0 instead of None. I've fixed this by changing the python code to use the scala None instead of the python None.
I've inspected manually to make sure that this works but since I am not aware of any python serialization -> scala deserialization tests in mleap, I've not added any unit tests to address this.


talalryz:
@lucagiovagnoli @voganrc for review

python/mleap/pyspark/feature/math_binary.py
```diff
@@ -57,15 +60,18 @@ def __init__(
         :param inputA: column name for the left side of operation (string)
         :param inputB: column name for the right side of operation (string)
         :param outputCol: output column name (string)
+        :param defaultA: Default to use instead of inputA
```
voganrc, line 63: So `inputA` represents the column name but `defaultA` is the value for that column? Can we make this more clear?

python/mleap/pyspark/feature/math_binary.py
```diff
@@ -102,7 +105,8 @@ def setParams(self, inputA=None, inputB=None, outputCol=None):
         """
         Sets params for this MathBinary.
         """
-        kwargs = self._input_kwargs
+        # For the correct behavior of MathBinary, params that are None must be unset
```
voganrc, line 108: What happens if they are not unset?

python/tests/pyspark/feature/math_binary_test.py
```diff
@@ -147,3 +146,39 @@ def test_serialize_deserialize_pipeline(self):
...
+        ], columns=['add(f1, f2)'])
+
+        result = add_transformer.transform(none_df).toPandas()[['add(f1, f2)']]
+        assert_frame_equal(expected_df, result)
+
+    def test_add_math_binary_default_inputA(self):
+        add_transformer = MathBinary(
+            operation=BinaryOperation.Multiply,
+            inputB="f2",
+            outputCol="add(f1, f2)",
```
voganrc, line 171: Shouldn't the outputCol be `multiply(f1, f2)`?

python/mleap/pyspark/feature/math_binary.py
talalryz: The spark hook for the transformer does a check for isSet(col1), which passes even when the col is None causing the transform to fail

python/mleap/pyspark/feature/math_binary.py
talalryz: Yes, I can make this more clear. Basically, when you provide defaultA, you should skip inputA. This is useful for cases where you want to e.g., add +1 to all entries in a col or mutliply a col by 3. 
I had earlier thought about making assertions to make sure that users provide either inputA or defaultA but not both, but thought against doing so to keep parity with the scala transformer

lucagiovagnoli:
@ancasarb it seems my MathBinary is currently broken on scala deserialization on `0.16.0`, do you think it's possible to publish `0.16.1` to include this fix ?

ancasarb:
hey @lucagiovagnoli sounds good, i'll release 0.16.1 early this week.

PR #708, opened 2020-07-22
Upgrade to xgboost 1.0.0 - Use h2oai Predictor

### LINKED ISSUE #697: Can we release scala 2.12 support in central maven repository
Currently I only see scala 2.11 support for mleap-xgboost-spark and mleap-xgboost-runtime in central maven repository, can you add release for scala 2.12, spark 2.4.5 has to use scala2.12

lucagiovagnoli:
Fixes #697 

## Major changes
- Use xgboost 1.0.0
- Adopt the [h2oai Predictor library](https://github.com/h2oai/xgboost-predictor). As I mentioned in a past comment [here](https://github.com/combust/mleap/pull/645#discussion_r416056178), the h2oai version is actively maintained, as opposed to the [komiya-atsushi](https://github.com/komiya-atsushi/xgboost-predictor-java/issues/21) version which seems completely abandoned.

## Minor
- Testing xgboost-runtime on scala 2.12! xgboost 1.0.0 is compatible with scala 2.12
- Adding a dev plugin dependency-graph, useful to print out the dependency graph (`sbt dependency-graph`)
- Interface fixes
- Change xgboost testing dataset (use diabetes rather than agaricus) because agaricus had missing columns and the xgboost libSVM loader would pollute logs with WARNINGs. Also use `indexing_mode=1`.

NOTE: this has been rebased on top of https://github.com/combust/mleap/pull/709

@talalryz @voganrc for review


lucagiovagnoli:
I was a bit nervous too but tests passed with both dataset (agaricus and diabetes). The problem are just those WARNINGs. 
So, `xgboost>=1.0.0` will print a WARNING to screen for each line containing a missing value.
```
WARNING: /xgboost/src/learner.cc:979: Number of columns does not match number of features in booster
```

I also commented here on [dmlc/xgboost](https://github.com/dmlc/xgboost/issues/5841#issuecomment-663182422) and they suggested to add `indexing_mode=1` to solve the problem of a dataset starting from index `1`. But libSVM still complains when there's a column missing from a dataset row. 


Example of the diabetes dataset:
```
0  1:6.000000 2:148.000000 3:72.000000 4:35.000000 5:0.000000 6:33.599998 7:0.627000 8:50.000000
+1  1:1.000000 2:85.000000 3:66.000000 4:29.000000 5:0.000000 6:26.600000 7:0.351000 8:31.000000
0  1:8.000000 2:183.000000 3:64.000000 4:0.000000 5:0.000000 6:23.299999 7:0.672000 8:32.000000
```
All of the columns from 1 to 8 are defined, while in the agaricus, not all  of the 127 features are always defined.

The problem is that the logs become too long, both locally and in Travis. I'm not sure I can silence all of those warnings, I also don't want to silence all warnings.  One way could be to load the data via a different method  but SVM is so convenient

What do you suggest?

jsleight:
Seems like the warning just come from loading the libsvm file, so maybe its ok -- we're just relying on the upstream packages to have good test coverage for their null handling.  Agree we don't want to just silence all the warnings.

lucagiovagnoli:
I think this issue was fixed 23 days ago here: https://github.com/dmlc/xgboost/pull/5856 ? So we'd need to wait for xgboost 1.1.1 to set the #features to 127

I'm now wondering if we'd see all of those warnings in prod too with 1.0.0 (there's a warning for each line with a missing value)

jsleight:
This only matters for reading libsvm format.  We could even look at changing the test setup here to not read the libsvm files and just create a train/test dataset in another way.  E.g., [how xgboost's internal tests do it](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifierSuite.scala)

lucagiovagnoli:
> This only matters for reading libsvm format. We could even look at changing the test setup here to not read the libsvm files and just create a train/test dataset in another way. E.g., [how xgboost's internal tests do it](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifierSuite.scala)

I fixed the WARNINGS by recreating agaricus in CSV and loading from CSV. I'd like to try out upgrading to 1.1.1 directly as suggested in this comment https://github.com/combust/mleap/issues/697#issuecomment-692347769 because 1.1.1 should also fix those warnings and maybe it's better to jump to 1.1.1 as it's supposed to be more stable? I can also push this out for now and tackle 1.1.1 later @jsleight

jsleight (approved):
If this is working, then I'd ship the 1.0.0 migration first and do a follow up to bump up to later versions in the future.

lucagiovagnoli:
This does work and tests now pass. I only had to add a 3MB csv file to git (which we might remove on 1.1.1) but eh, it's not too much I think.

@ancasarb for final review

project/plugins.sbt
```diff
@@ -8,5 +8,7 @@ addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.6")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.6.1")
 addSbtPlugin("com.typesafe.sbt" % "sbt-git" % "0.9.3")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.7")
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
```
ancasarb, line 11: any way we can depend not on a RC?

.travis.yml
```diff
@@ -2,6 +2,14 @@
 sudo: required
 dist: trusty
 
+addons:
```
ancasarb, line 5: can you please remind me why these addons are needed?

Makefile
```diff
@@ -11,10 +11,10 @@ test_benchmark:
 	sbt "+ mleap-benchmark/test"
 
 test_xgboost_runtime:
-	sbt "mleap-xgboost-runtime/test"
+	sbt "+ mleap-xgboost-runtime/test"
 
 test_xgboost_spark:
-	sbt "mleap-xgboost-spark/test"
+	sbt "+ mleap-xgboost-spark/test"
```
ancasarb, line 17: We can remove these separate tasks if we now add the `mleap-xgboost-runtime` and `mleap-xgboost-spark` subprojects in the `aggregatedProjects` list in MleapProject.scala`.

travis/travis_publish.sh
```diff
@@ -4,8 +4,8 @@ if [[ $TRAVIS_BRANCH == 'master' ]] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; the
   source travis/extract.sh
   source travis/docker.sh
   sbt "+ publishSigned" \
-      "mleap-xgboost-runtime/publishSigned" \
-      "mleap-xgboost-spark/publishSigned" \
+      "+ mleap-xgboost-runtime/publishSigned" \
+      "+ mleap-xgboost-spark/publishSigned" \
```
ancasarb, line 8: same as above, these can be removed if we add the mleap-xgboost-runtime and mleap-xgboost-spark subprojects in the aggregatedProjects list in MleapProject.scala`.

project/Dependencies.scala
```diff
@@ -67,10 +68,12 @@ object Dependencies {
...
     )
-    val xgboostDep = "ml.dmlc" % "xgboost4j" % xgboostVersion // scala 2.11 only
-    val xgboostPredictorDep = "biz.k11i" % "xgboost-predictor" % "0.3.1"
 
-    val xgboostSparkDep = "ml.dmlc" % "xgboost4j-spark" % xgboostVersion // scala 2.11 only
+    val kryo = "com.esotericsoftware" % "kryo" % kryoVersion
+    val xgboostDep = "ml.dmlc" %% "xgboost4j" % xgboostVersion exclude("com.esotericsoftware.kryo", "kryo")
+    val xgboostPredictorDep = "ai.h2o" % "xgboost-predictor" % "0.3.17" exclude("com.esotericsoftware.kryo", "kryo")
+
+    val xgboostSparkDep = "ml.dmlc" %% "xgboost4j-spark" % xgboostVersion exclude("com.esotericsoftware.kryo", "kryo")
```
ancasarb, line 76: did we run into issues with the kyro versions?

ancasarb (changes requested):
Looks great overall, thank you! Can you please just make the small change to simplify the build scripts? Thanks again!

project/Dependencies.scala
lucagiovagnoli: They are brought in elsewhere and there is a version conflict. This is just making sure that others are chosen

project/plugins.sbt
lucagiovagnoli: This dependency graph is only used from command line during development to print out the various jars but I can definitely downgrade to the latest stable

Makefile
lucagiovagnoli: That makes sense. What was the original reason for them to be separated ? I assume because xgboost wouldn't build on 2.12 ?

.travis.yml
lucagiovagnoli: Uh I wasn't able to build it without it. I can post the error message here by removing them and running tests again if you'd like. I also noticed that it was necessary in the spark-3.0.0 release branch here: https://github.com/combust/mleap/blob/spark-3.0.0/.travis.yml#L14

ancasarb: 👍

ancasarb (approved):

PR #709, opened 2020-07-27
Better parallelization of scala tests

lucagiovagnoli:
I'm trying to get tests to run quicker by parallelizing most of the scala steps.

I think that I've hit some Travis timeout in this PR: https://github.com/combust/mleap/pull/708/files  and I'm trying to fix that. But also, who doesn't like faster test execution? :)

@talalryz for review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,11 @@ jobs:
       python: 3.6
       install:
         - pip install tox
+      before_script:
+        - >
+          curl
+          --create-dirs -L -o $HOME/.sbt/launchers/0.13.18/sbt-launch.jar
+          https://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/0.13.18/sbt-launch.jar
       script:
         - make py36_test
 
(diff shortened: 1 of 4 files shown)


lucagiovagnoli:
Example UI on master (w/ deploy stage) 
Screen Shot 2020-07-27 at 16 00 38

talalryz:
this lgtm! I am surprised that the old shell scripts was running sbt test + tests for each package individually.
Do you know if there was any reason for that?

lucagiovagnoli:
> this lgtm! I am surprised that the old shell scripts was running sbt test + tests for each package individually.
> Do you know if there was any reason for that?

Yeah `sbt +test` is not actually running all of the others (xgboost, benchmark etc.), so we need to keep that in.

lucagiovagnoli:
@ancasarb do you think we can merge this? It's a nice quality of life improvement on test iteration speed

lucagiovagnoli:
I've noticed these tests failing in #708 , so I reran them here without changing the code. Apparently something with travis changed cause these were passing and I haven't changed anything

ancasarb (approved):
lgtm! thank you!

PR #717, opened 2020-09-16
Create proper temporary directories

### LINKED ISSUE #715: Don't hardcode temporary path in tests
We're having issues where people cannot run mleap tests on the same machine at the same time, because the testing path for serialized models [here](https://github.com/combust/mleap/blob/41dbde99e389873fc609083cce5d610cea9e9170/mleap-spark-testkit/src/main/scala/org/apache/spark/ml/parity/SparkParityBase.scala#L81) is hardcoded. (Issues with concurrent access)

lucagiovagnoli:
Fixes #715 

Rebased on top of https://github.com/combust/mleap/pull/709


lucagiovagnoli:
@talalryz and @ancasarb  for review (only last commit since this is rebased on top of #709

ancasarb (approved):
lgtm!

PR #719, opened 2020-10-02
Optimize Interaction and CountVectorizer

talalryz:
### Interaction

This pr includes a performance optimization for the `InteractionModel`. 
The current version makes a roundtrip conversion; first `mleapTensorToSparkVector` and then`sparkVectorToMleapTensor` when making transformations in mleap runtime. This is a very expensive operation and can be avoided by completely avoiding these casts.

In this pr, I've:

- added cases to `foreachNonzeroOutput` for `DenseTensor` and `SparseTensor` which avoids the first cast from a `Tensor` to `Vector`. 

- changed `apply` to return an MLeap tensor instead of a spark Vector. Specifically, it now returns a `Tensor[Double]` instead of the `Vector` returned by the regular `apply`, and consequently avoids the second cast from `Vector` to `Tensor` inside the `Interaction` transformer.

## CountVectorizer

This pr also includes a change to `CountVectorizerModel`. It contains a cast from `sparkVectorToMleapTensor` which is again expensive and can be avoided by cleverly instantiating the `SparseTensor` ourselves.  The return type of `CountVectorizerModel.apply` is now a `Tensor[Double]` instead of `Vector`.

When instantiating the SparseTensor, we use the knowledge that we know the output size of the tensor at model initialization to score a cache of `Seq(Int)`. These seq are required by the `SparseTensor` api for the indices. However, initializing these at runtime is pretty expensive, and using the seq cache avoids this (The same trick is also used in the `InteractionModel`)

## Testing

I've added tests for the newly added code. These tests pass.


mleap-core/src/main/scala/ml/combust/mleap/core/feature/InteractionModel.scala
```diff
@@ -18,9 +18,28 @@ case class InteractionModel(featuresSpec: Array[Array[Int]],
...
+    for (i <- 0 to outputSize){
+      arr += Seq(i)
+    }
+    arr.result()
+  }
   val encoders: Array[FeatureEncoder] = featuresSpec.map(FeatureEncoder.apply)
 
   def apply(features: Seq[Any]): Vector = {
+    val (size, indices, values) = _apply(features)
+    Vectors.sparse(size, indices, values)
```
jsleight, line 34: `.compressed`

jsleight (commented):
lgtm!

We probably want this style of optimization for the other transformers we plan to use as well.  Definitely `CountVectorizerModel`, maybe also `OneHotEncoderModel`, `BinarizerModel`, `BucketizerModel`, and `StandardScalarModel` too? Can obviously break things up into seperate PRs though.

mleap-core/src/main/scala/ml/combust/mleap/core/feature/InteractionModel.scala
talalryz: added `.compressed`

talalryz:
> lgtm!
> 
> We probably want this style of optimization for the other transformers we plan to use as well. Definitely `CountVectorizerModel`, maybe also `OneHotEncoderModel`, `BinarizerModel`, `BucketizerModel`, and `StandardScalarModel` too? Can obviously break things up into seperate PRs though.

I've included changes to the `CountVectorizerModel` here as well since it was pretty straightforward. I can take a look at the others as well and include them here if they are straightforward like CV but i'll tackle them in a separate pr if they are a little more involved

talalryz:
@ancasarb @lucagiovagnoli  for review

jsleight (approved):

mleap-core/src/main/scala/ml/combust/mleap/core/feature/CountVectorizerModel.scala
```diff
@@ -34,8 +44,16 @@ case class CountVectorizerModel(vocabulary: Array[String],
...
     }
+    effectiveCounts
+  }
+
+  def apply(document: Seq[String]): Vector = {
+    Vectors.sparse(outputSize, _apply(document))
+  }
 
-    Vectors.sparse(dict.size, effectiveCounts)
+  def mleapApply(document: Seq[String]): Tensor[Double] = {
```
lucagiovagnoli, line 54: I'm not sure I understand the naming. `apply` vs `mleapApply` ? 
One is the legacy Vectors.sparse `apply` and the other is the new one, should these be called with more informative names ?

talalryz: The old apply can just be dropped. Earlier i had incorrectly assumed that `CountVectorizerModel.apply`  is called by both mleap-runtime and spark. However, since this code is only called by  mleap runtime which sends expects mleap tensors, i can drop the legacy apply completely and just use the new one.

I can do the same change for Interaction

talalryz: I've updated the code to remove the legacy apply and just use `Tensor` instead of `Vector` in `InteractionModel` and `CountVectorizerModel`. We can do this since both of these transformers are being used only by the corresponding transformers `CountVectorizer` and `Interaction` in mleap runtime which expect `Tensor` outputs.

mleap-core/src/main/scala/ml/combust/mleap/core/feature/CountVectorizerModel.scala
```diff
@@ -27,15 +36,34 @@ case class CountVectorizerModel(vocabulary: Array[String],
...
+        }
+      }
+    }else{
+      while (iterator.hasNext){
+        val (termIndex, count) = iterator.next()
+        if (count>= effectiveMinTF){
+          valuesArray += count
+          indicesArray += seqCache(termIndex)
+        }
+      }
```
talalryz, line 58: while loops are faster than filter + map. This also eliminates the need to create a Seq of tuples and then unzip or iterate over it to get indices and values

jsleight (approved):

ancasarb (approved):
looks good!

PR #720, opened 2020-10-09
Fix broken build

lucagiovagnoli:
Fixing scala.reflect.internal.MissingRequirementError: object java.lang.Object in compiler mirror not found.

Seems to be similar to
https://github.com/sbt/zinc/issues/641

EDIT: tested that it `publishLocal` passed  here: https://travis-ci.org/github/combust/mleap/jobs/734346400

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ services:
 
 language: scala
 scala:
-  - 2.11.8
+  - 2.11.12
 jdk:
   - oraclejdk8  # sbt +compile fails on jdk11 (xenial's default)
 


ancasarb (approved):

.travis.yml
```diff
@@ -65,7 +65,7 @@ jobs:
       script:
         - make py37_test
 
-    - if: (NOT type IN (pull_request)) AND (branch = master)
+    - if: (NOT type IN (pull_request))
```
lucagiovagnoli, line 68: I'll revert this back after verifying that this works

PR #724, opened 2020-10-27
U/talal/add support xgboost predictor regression

talalryz:
## Description

This adds support for using xgboost predictor when using xgboost regressor. Currently, if you serialize an xgboost regressor,  it will always be deserialized as `dmlc.xgboost` and there is no option to use `biz.k11i.xgboost`. 

This pr picks up from https://github.com/combust/mleap/pull/645, which only implemented xgboost predictor analogs for xgboost classifier and not xgboost regressor.
Some  things to note:
- `XGBoostPredictorRegression` will only generate one output Col (`prediction`)  for performance reasons (similar to `XGBoostPredictorClassifation` )
- `XGBoostPredictorRegression` has an output of type `Double` to match the behavior of `XGBoostRegression`

## Tests
A new test suite `XGBoostPredictorRegressionModelParitySpec`, to test the new code and its parity with the existing dmlc `XGBoostRegressor`. tests pass


mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorRegressionModelParitySpec.scala
```diff
@@ -0,0 +1,109 @@
...
+  it("A deserialized XGBoost4j has the same results of a deserialized Predictor"){
+    val xgboost4jTransformer = trainRegressor
+
+    val mleapBundle = serializeModelToMleapBundle(xgboost4jTransformer)
+
+    val deserializedXGBoost4jTransformer: Transformer = loadMleapTransformerFromBundle(mleapBundle)
+    val deserializedXGBoost4jResult = deserializedXGBoost4jTransformer.transform(leapFrameBinomial).get
+
+    val deserializedPredictorTransformer: Transformer = loadXGBoostPredictorFromBundle(mleapBundle)
+    val deserializedPredictorResult = deserializedPredictorTransformer.transform(leapFrameBinomial).get
```
voganrc, line 75: Maybe add extra sanity checks here to assert that one is an instance of `XGBoost4j` and the other is `XGBoostPredictor`

voganrc (approved):
LGTM!

jsleight (approved):

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorRegressionModelParitySpec.scala
talalryz: I've added extra sanity checks here.
Thank you for the feedback!

ancasarb (approved):
looks great, thank you!

PR #735, opened 2021-01-25
remove unneccessary headers for GET requests without payloads

jwgwalton:
Remove the requirement to have Content type headers on GET requests with empty payloads.


ancasarb (approved):
lgtm, thank you!

Issue #721, opened 2020-10-09
Deserialization of RandomForestRegression model into a spark stage is incorrect if the number of trees is set to a value other than the default


leahmcguire:
For the spark RandomForestRegressor model the deserialized spark model will always have the default number of trees (20) resulting in incorrect scores on deserialization as a spark stage. (It produces the correct scores when deserialized as an mleap stage)

ancasarb:
thanks for raising this bug @leahmcguire , i've created a PR to fix this and it'll be included in the next release. thanks again!

PR #736, opened 2021-01-28
fixes issue #721

ancasarb:


talalryz (approved):
This looks good to me. I had one small question, do we not need to make changes to [this op](https://github.com/combust/mleap/blob/master/mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/classification/RandomForestClassifierOp.scala),

ancasarb:
thanks @talalryz , on mleap side, we don't keep track of the number of trees separately (instead it's just a list of trees), so that's why that op hasn't changed.

PR #741, opened 2021-02-17
annotate SerializationFormat as Serializable

clementemmanuel:
If there's no introduced complications from doing this, this would help simplify the resolution to spark task serialization issues we're facing when trying to broadcast mleap models in our inference pattern.


ancasarb (approved):
looks good to me, it doesn't seem to break anything so i'll merge it in.

PR #746, opened 2021-04-07
fix test failure with gensim 4.x

zhxiaogg:
The latest release gensim 4.x has API breaks, see [Migrating from Gensim 3.x to 4](https://github.com/RaRe-Technologies/gensim/wiki/Migrating-from-Gensim-3.x-to-4)


ancasarb (approved):
looks good, thanks for the fix!

PR #753, opened 2021-05-06
Spark 3.1.1

emitc2h:
This branch provides the following dependency upgrades:
- scala 2.11 -> 2.12
- spark 2.4.5 -> 3.1.1
- xgboost 1.0.0 -> 1.3.1
- akka 2.5.12 -> 2.6.14
- akka HTTP 10.0.3 -> 10.2.4
- hadoop 2.6.5 -> 2.7.4

### DIFF
--- a/mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/classification/RandomForestClassifierOp.scala
+++ b/mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/classification/RandomForestClassifierOp.scala
@@ -56,7 +56,7 @@ class RandomForestClassifierOp extends SimpleSparkOp[RandomForestClassificationM
         numFeatures = numFeatures,
         numClasses = numClasses,
         _trees = models)
-      m.setNumTrees(models.size)
+      m.set(m.getParam("numTrees"), models.size)
       model.getValue("thresholds").
         map(t => m.setThresholds(t.getDoubleList.toArray)).
         getOrElse(m)
@@ -69,7 +69,7 @@ class RandomForestClassifierOp extends SimpleSparkOp[RandomForestClassificationM
       numFeatures = model.numFeatures,
       numClasses = model.numClasses)
     if (model.isDefined(model.thresholds)) { r.setThresholds(model.getThresholds) }
-    if (model.isDefined(model.numTrees)) { r.setNumTrees(model.getNumTrees) }
+    if (model.isDefined(model.numTrees)) { r.set(r.getParam("numTrees"), model.getNumTrees) }
     r
   }
 
--- a/mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModelParitySpec.scala
+++ b/mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModelParitySpec.scala
@@ -38,7 +38,8 @@ class XGBoostPredictorClassificationModelParitySpec extends FunSpec
         assert(
           almostEqualSequences(
             Seq(boosterProbability.values),
-            Seq(mleapResult.dataset.head.getTensor[Double](mleapProbabilityColIndex).toArray)
+            Seq(mleapResult.dataset.head.getTensor[Double](mleapProbabilityColIndex).toArray),
+            precision = 1e-6
           )
         )
     }
@@ -58,7 +59,9 @@ class XGBoostPredictorClassificationModelParitySpec extends FunSpec
         assert(
           almostEqualSequences(
             Seq(row1.getTensor[Double](probabilityIndex).toArray),
-            Seq(row2.getTensor[Double](probabilityIndex).toArray)))
+            Seq(row2.getTensor[Double](probabilityIndex).toArray),
+            precision = 1e-6
+          ))
       }
     }
   }
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -16,7 +16,7 @@ object Common {
   lazy val defaultSettings = buildSettings ++ sonatypeSettings
 
   lazy val buildSettings: Seq[Def.Setting[_]] = Seq(
-    scalaVersion := "2.11.12",
+    scalaVersion := "2.12.10",
     crossScalaVersions := Seq("2.11.12", "2.12.10"),
     scalacOptions ++= Seq("-unchecked", "-deprecation", "-feature"),
     fork in Test := true,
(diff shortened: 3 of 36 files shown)


project/Common.scala
```diff
@@ -16,8 +16,8 @@ object Common {
   lazy val defaultSettings = buildSettings ++ sonatypeSettings
 
   lazy val buildSettings: Seq[Def.Setting[_]] = Seq(
-    scalaVersion := "2.11.12",
-    crossScalaVersions := Seq("2.11.12", "2.12.10"),
+    scalaVersion := "2.12.10",
```
jsleight, line 19: want to try `2.12.13` as latest 2.12 series?  I'd guess it is a no-op change

project/Common.scala
```diff
@@ -28,7 +28,7 @@ object Common {
       if(isSnapshot.value) {
         Seq(
           "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots",
-          "ASF Snapshots" at "https://repository.apache.org/content/groups/snapshots"
+          "ASF spark" at "https://repository.apache.org/content/repositories/orgapachespark-1341" // Apache spark repo
```
jsleight, line 31: `https://repository.apache.org/content/repositories/orgapachespark-1341` gives me a 404.  What is this change trying to do?

python/requirements-dev.txt
```diff
@@ -1,6 +1,7 @@
+-f https://dist.apache.org/repos/dist/dev/spark/v3.0.0-rc1-bin/
```
jsleight, line 1: why find-links from here?

1. it is not the same spark version as we're installing
2. pip will bring pyspark dependencies already

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModelParitySpec.scala
```diff
@@ -38,7 +38,8 @@ class XGBoostPredictorClassificationModelParitySpec extends FunSpec
         assert(
           almostEqualSequences(
             Seq(boosterProbability.values),
-            Seq(mleapResult.dataset.head.getTensor[Double](mleapProbabilityColIndex).toArray)
+            Seq(mleapResult.dataset.head.getTensor[Double](mleapProbabilityColIndex).toArray),
+            precision = 1e-6
```
jsleight, line 42: needing to specify the precision level makes me nervous.  iirc, xgboost is always 32-bit floats, so using double's here might be the issue?  Or the predictor library also needs to be upgraded for the latest xgboost version?

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/classification/RandomForestClassifierOp.scala
```diff
@@ -69,7 +69,7 @@ class RandomForestClassifierOp extends SimpleSparkOp[RandomForestClassificationM
       numFeatures = model.numFeatures,
       numClasses = model.numClasses)
     if (model.isDefined(model.thresholds)) { r.setThresholds(model.getThresholds) }
-    if (model.isDefined(model.numTrees)) { r.setNumTrees(model.getNumTrees) }
+    if (model.isDefined(model.numTrees)) { r.set(r.getParam("numTrees"), model.getNumTrees) }
```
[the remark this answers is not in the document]
emitc2h: Ya, I wish it was easier to track down the logic behind those changes.

project/Common.scala
emitc2h: Let me try that.

project/Common.scala
emitc2h: I'm inheriting this change from the original spark-3.0.0 branch, @mengxr, do you remember what this is about?

python/requirements-dev.txt
emitc2h: I'm also inheriting this from the original spark-3.0.0 branch. I don't mind dropping this unless @mengxr had a good reason.

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModelParitySpec.scala
emitc2h: I see, this is helpful. Let me see if I can get rid of this.

python/requirements-dev.txt
emitc2h: I can at least set it to spark 3.1.1.

mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModelParitySpec.scala
emitc2h: I dug further into this, and I'm hoping you can help me make sense of what I'm seeing. First, trying to use floats instead of doubles here is rather involved since the reason Doubles are used in the first place is so the code plays well with data structures such as Spark ML Vectors, for which there is no Float-specific implementation (only Double).

The test that fails that I'm trying to fix is [this one](https://github.com/combust/mleap/blob/61c15aa128fe06a2096b09007ad6f047c71fbb6f/mleap-xgboost-runtime/src/test/scala/ml/combust/mleap/xgboost/runtime/XGBoostPredictorClassificationModelParitySpec.scala#L117). I've noticed by fooling around that if I load the transformer using `loadMleapTransformerFromBundle` instead of `loadXGBoostPredictorFromBundle`, the error goes away and the predictions match exactly. Can you help me understand the difference between those two?

jsleight: The difference between `loadMleapTransformerFromBundle` and `loadXGBoostPredictorFromBundle` is which library is used in the backend.  There are two libraries for making predictions for xgboost models:

- The dmlc xgboost library.  This is the official open source code for xgboost, and its java/scala API is just a JNI wrapper around xgboost's main c++ code.  The JNI process introduces extra overhead and makes predictions for single rows be slow.
- The xgboost-predictor library is a java implementation that only allows for making predictions (i.e., you can't train an xgboost model with this library).  This library has been benchmarked as being several orders of magnitude lower latency for making predictions on single rows.

mleap has a special op so that folks can decide to use the predictor library instead of the dmlc xgboost.  Your observation that the predictions match exactly when using the loadMleapTransformerFromBundle (i.e., dmlc xgboost) but don't match when using loadXGBoostPredictorFromBundle (i.e., xgboost-predictor library) indicates that there is a problem which we need to fix.

PR #760, opened 2021-06-03
Add Tensorflow 2.0 with new tensorflow/java binding

austinzh:
Use https://github.com/tensorflow/java instead of
https://github.com/tensorflow/tensorflow

Use org.tensorflow.tensorflow-core-platform 0.3.1 to match Tensorflow 2.4.1


mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+
+    it("vector") {
+      val array = Array(20, 10, 5)
+      val ml_tensor = Tensor.denseVector(array)
+      val tf_tensor_converted = MleapConverter.convert(ml_tensor, TensorType.Int())
+      val ml_tensor_converted = TensorflowConverter.convert(tf_tensor_converted, TensorType.Int())
+      assert(ml_tensor.dimensions == ml_tensor_converted.dimensions)
+      assert(ml_tensor.base == ml_tensor_converted.base)
+      assert(ml_tensor.size == ml_tensor.size)
+      assert(ml_tensor.toArray sameElements ml_tensor_converted.toArray)
```
jsleight, line 37: could consider putting these 4 lines as a helper method since they get repeated so often.  Or even adding an equals method onto mleap tensor so you can just `assert(ml_tensor == ml_tensor_converted)`

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+      assert(
+        Helper.scalars(shape).map(indices => tf_tensor_converted.getObject(indices.map(_.toLong): _*))
+          sameElements Helper.scalars(shape).map(indices => ml_tensor(indices: _*))
+      )
+    }
+
+    it("byte string nd dense tensors") {
+      val random = new java.security.SecureRandom
+      //      val shape = Seq(3, 3, 3, 3, 3, 3, 3)
+      val shape = Seq(3, 3)
```
jsleight, line 148: why this was needed? (i'm mostly just curious and don't really expect any code changes)

mleap-databricks-runtime-testkit/src/main/scala/ml/combust/mleap/databricks/runtime/testkit/Boot.scala
```diff
@@ -1,17 +1,17 @@
...
+//    master("local[2]").
+//    getOrCreate()
+//  val sqlContext = session.sqlContext
+//
+//  new TestSparkMl(session).run()
+//  new TestTensorflow(session).run()
+//  new TestXgboost(session).run()
+//
+//  session.close()
+//}
```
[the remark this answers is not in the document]
austinzh: I change this to WIP cause I found out it still  needs some work.  All databricks related test will fail, I am fixing it.

project/Dependencies.scala
```diff
@@ -15,7 +16,7 @@ object Dependencies {
   lazy val loggingVersion = "3.9.0"
   lazy val slf4jVersion = "1.7.25"
   lazy val awsSdkVersion = "1.11.349"
-  val tensorflowVersion = "1.11.0"
+  val tensorflowVersion = "0.3.1"
```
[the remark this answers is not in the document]
austinzh: Sure

project/Dependencies.scala
```diff
@@ -38,8 +39,7 @@ object Dependencies {
     val jTransform = "com.github.rwl" % "jtransforms" % "2.4.0" exclude("junit", "junit")
     val commonsIo = "commons-io" % "commons-io" % "2.5"
     val tensorflowDeps = Seq(
-      "org.tensorflow" % "libtensorflow" % tensorflowVersion,
-      "org.tensorflow" % "libtensorflow_jni" % tensorflowVersion
+      "org.tensorflow" % "tensorflow-core-platform" % tensorflowVersion
```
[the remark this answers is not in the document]
austinzh: Yes. We should be able can use the JavaCPP to only download the needed  native libraries.

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
austinzh: I forget to change back to val shape = Seq(3, 3, 3, 3, 3, 3, 3), I use simpler shape for debug

jsleight (approved):
lgtm, exciting to see newer TF support 🎉

mleap-tensor/src/main/scala/ml/combust/mleap/tensor/Tensor.scala
```diff
@@ -66,30 +75,12 @@ case class DenseTensor[T](values: Array[T],
...
-      var tn = tail.head
-      tail = tail.tail
-      while(ti < dimensions.size) {
-        tn *= dimensions(ti)
-        ti += 1
-      }
-      dimI += 1
-      i += 1
-      n += tn
-    }
```
voganrc, line 84: After reading this [blog post](https://eli.thegreenplace.net/2015/memory-layout-of-multi-dimensional-arrays), your new code makes sense to me:

Each time you increment a particular array index, you must traverse the memory addresses for all the faster-changing dimensions that come in between. It makes sense to me that the length of this traversal is equal to the product of all the faster-changing dimension sizes. Since you initialize `n = index.last`, it seems that the final dimension must be the fastest changing one, making your code row-major.

The old code confuses me though... Since they initialize `n = indices.head`, it would seem that the first dimension is the fastest-changing one, making it column-major. But the loop that looks like this:
```
while(ti < dimensions.size) {
    tn *= dimensions(ti)
    ti += 1
}
```
Seems to be multiplying `tn` by all the slower-changing dimensions that come after it. Instead, I think it should be multiplying together all the faster-changing dimensions that come **before** it.

Is the legacy code incorrect?

project/Dependencies.scala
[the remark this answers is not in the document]
austinzh: introduced a environment variable TENSORFLOW_PLATFORMS to determine which platform should support.
By default we download linux, windows, macosx.
But user can always build a jar with only linux or with mkl or gpu support.

The reason not using java cpp is sbt only have sbt-javacpp to use javacpp presets, but presets only support tensorflow 1

mleap-tensor/src/main/scala/ml/combust/mleap/tensor/Tensor.scala
austinzh: Yes. The legacy code has bug.
Here is the code the verify.
describe("Test Legacy DenseTensor Get") {
  def legacy_get(indices: Seq[Int], dimensions: Seq[Int]) = {
    //This is the legacy get method from DenseTensor
    var i = 0
    var dimI = 1
    var n = indices.head
    var tail = indices.tail
    while (i < tail.size) {
      var ti = dimI
      var tn = tail.head
      tail = tail.tail
      while (ti < dimensions.size) {
        tn *= dimensions(ti)
        ti += 1
      }
      println(tn)
      dimI += 1
      i += 1
      n += tn
      println(n)
    }
    n
  }
  it("Fortran layout") {
    assert(legacy_get(Seq(0, 0), Seq(3, 4)) == 0)
    assert(legacy_get(Seq(2 , 3), Seq(3, 4)) == 3*3+2)
    assert(legacy_get(Seq(1, 1), Seq(3, 4)) == 1*3+1)
    assert(legacy_get(Seq(1, 1, 1), Seq(3, 4, 5)) == 1*4*3 + 1*3+ 1)
    assert(legacy_get(Seq(2, 3, 4), Seq(3, 4, 5)) == 4*4*3 + 3*3+ 2)
  }
}

mleap-tensor/src/test/scala/ml/combust/mleap/tensor/TensorSpec.scala
```diff
@@ -36,21 +83,61 @@ class TensorSpec extends FunSpec {
...
+  describe("DenseTensor") {
+    val random = new scala.util.Random
+    val shape = Seq(5, 4, 3, 2, 1)
+    val dims = shape.length
+    val size = shape.product
+    val floatArray = Array.fill(size) {
+      random.nextFloat
+    }
+
+    it("should return true for dense tensors with same elements and dimensions") {
```
voganrc, line 107: nit: this doesn't mention you are checking equality

mleap-tensor/src/test/scala/ml/combust/mleap/tensor/TensorSpec.scala
```diff
@@ -36,21 +83,61 @@ class TensorSpec extends FunSpec {
...
+
+    it("should return true for dense tensors with dimension -1 and same elements") {
+      val tensor1 = Tensor.create(floatArray, Seq(-1) ++ shape.slice(1, 5))
+      val tensor2 = Tensor.create(floatArray, shape.slice(0, 4) ++ Seq(-1))
+      assert(tensor1 == tensor2)
+    }
+
+    it("should return true for dense tensors each elements compare to original array") {
+      val tensor = Tensor.create(floatArray, dimensions = shape)
+      println(floatArray.mkString(" "))
```
voganrc, line 133: remove print statements

mleap-tensor/src/test/scala/ml/combust/mleap/tensor/TensorSpec.scala
```diff
@@ -80,16 +167,65 @@ class TensorSpec extends FunSpec {
...
+      assert(tensor1 != tensor2)
+    }
+
+    it("should return false for sparse tensors with same dimension but different elements") {
+      val tensor1 = Tensor.create(array, shape, Some(indices))
+      val tensor2 = Tensor.create(array.map(x => x * 2), shape, Some(indices))
       assert(tensor1 != tensor2)
     }
+
+    it("should raise exception for sparse tensors with dimension -1") {
```
voganrc, line 211: I think this duplicates a test that we have above

mleap-tensor/src/test/scala/ml/combust/mleap/tensor/TensorSpec.scala
```diff
@@ -80,16 +167,65 @@ class TensorSpec extends FunSpec {
...
+      assertThrows[IllegalArgumentException] {
+        Tensor.create(array, Seq(-1), Some(indices))
+      }
+    }
+
+    it("should return true for sparse tensors each elements compare to original array") {
+      val tensor = Tensor.create(array, shape, Some(indices))
+      println(array.mkString(" "))
```
voganrc, line 219: remove print statements

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/TensorflowTransformerSpec.scala
```diff
@@ -49,13 +49,17 @@ class TensorflowTransformerSpec extends FunSpec {
...
 
     val graphBytes = Files.readAllBytes(Paths.get(getClass.getClassLoader.getResource("optimized_wine_quality.pb").getPath))
-    val graph = new org.tensorflow.Graph()
-    graph.importGraphDef(graphBytes)
+    val graph = new Graph()
+    val graphDef = GraphDef.parseFrom(graphBytes)
+
+    graph.importGraphDef(graphDef)
+    val gb = graph.toGraphDef.toByteArray
+    println(gb.length)
```
voganrc, line 57: remove print statement

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
```diff
@@ -0,0 +1,155 @@
...
+import org.tensorflow.types.TInt64
+import org.tensorflow.types.TString
+import org.tensorflow.types.TUint8
+
+class MleapConverterSpec extends FunSpec {
+  def toIndices(dimensions: Traversable[Int]): Seq[Seq[Int]] = {
+    combine(dimensions.map(d => 0 until d))
+  }
+
+  def combine[A](xs: Traversable[Traversable[A]]): Seq[Seq[A]] =
```
voganrc, line 18: This seems to repeat source code... can we just import the original?

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
```diff
@@ -0,0 +1,155 @@
...
+    xs.foldLeft(Seq(Seq.empty[A])) {
+      (x, y) => for (a <- x.view; b <- y) yield a :+ b
+    }
+
+  def assertSameTensor(lhr: Tensor[_], rhs: Tensor[_]) = {
```
voganrc, line 23: Also need to check indices if Sparse

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
```diff
@@ -0,0 +1,155 @@
...
+      val tf_tensor_converted = MleapConverter.convert(ml_tensor, TensorType.ByteString()).asInstanceOf[TString]
+      val ml_tensor_converted = TensorflowConverter.convert(tf_tensor_converted, TensorType.ByteString()).asInstanceOf[Tensor[ByteString]]
+      assertSameTensor(ml_tensor, ml_tensor_converted)
+      // make sure same indices has same value between tensorflow tensor and mleap tensor
+      val expected = toIndices(shape).map(indices => ml_tensor(indices: _*).bytes)
+      val actual = toIndices(shape).map(indices => tf_tensor_converted.asBytes().getObject(indices.map(_.toLong): _*))
+      assert(expected.zip(actual).forall({ case (lhs, rhs) => lhs sameElements rhs }))
+    }
+
+    it("string nd dense tensors") {
```
voganrc, line 141: We seem to test converting Dense tensors only... can we test Sparse as well?

project/Dependencies.scala
```diff
@@ -15,10 +15,12 @@ object Dependencies {
   lazy val loggingVersion = "3.9.0"
   lazy val slf4jVersion = "1.7.25"
   lazy val awsSdkVersion = "1.11.349"
-  val tensorflowVersion = "1.11.0"
+  val tensorflowVersion = "0.3.1" // Match Tensorflow 2.4.1 https://github.com/tensorflow/java/#tensorflow-version-support
```
voganrc, line 18: Hmm maybe we should rename this to `tensorflowJavaVersion` to make it more clear that it's a different package

mleap-tensor/src/main/scala/ml/combust/mleap/tensor/Tensor.scala
voganrc: @ancasarb This is the main part that we need you to verify
If we're correct, then everybody's `DenseTensor` indexing has been wrong this entire time

That would be very surprising

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
austinzh: We convert every Mleap Tensor to DenseTensor at the beginning of `MleapConverter.convert`.
But this bring me to a question.
Currently for MLeap SparseTensor, we call `toDense` in `MleapConverter.convert` to covert it to DenseTensor before convert it to Tensorflow Tensor.
This approach will introduce memory problem if the Sparse Tensor is really sparse.
Tensorflow has SparseTensor concept as well.  I am thinking of convert to Tensorflow SparseTensor if it's a MLeap SparseTensor.
@jsleight @ancasarb

jsleight: Respecting the sparse vs dense makes sense to me.  Exactly as you mentioned, could be very memory inefficient as is.

mleap-tensor/src/main/scala/ml/combust/mleap/tensor/Tensor.scala
```diff
@@ -149,29 +161,10 @@ case class SparseTensor[T](indices: Seq[Seq[Int]],
...
-        if (values.isEmpty) {
-          if (obj.values.isEmpty) { true }
-          else { false }
-        } else if(indices.length == obj.indices.length && indices == obj.indices) {
-          ((dimensions == obj.dimensions) ||
-            ((dimensions.head == -1 || obj.dimensions.head == -1) && dimensions.tail == obj.dimensions.tail)) &&
-              (values sameElements obj.asInstanceOf[SparseTensor[T]].values)
-        } else { false }
-      } else { false }
+        indices == obj.indices &&
```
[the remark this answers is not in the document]
austinzh: I change it to the sealed trait Tensor[T] equals method. 
And override it in Dense and Sparse to use it.
Then user don't have any chance to make mistake.

mleap-tensor/src/test/scala/ml/combust/mleap/tensor/TensorSpec.scala
austinzh: one is vector, the other one is Tensor
Because the indices is different

mleap-tensorflow/src/main/scala/ml/combust/mleap/tensorflow/converter/TensorflowConverter.scala
```diff
@@ -1,42 +1,86 @@
...
+    val sb = new StringBuilder
+    for (b <- bytes) {
+      sb.append(String.format("%02x ", Byte.box(b)))
     }
+    sb.toString
   }
-}
+    def convert(tensor: tensorflow.Tensor, tensorType: TensorType): DenseTensor[_] = {
+      val dimensions = getShape(tensor)
+      val size = getSize(tensor)
```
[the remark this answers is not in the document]
austinzh: I test with empty Tensor converted back and forward, it work.
@ancasarb @jsleight any thought about this `max(1, size)` ?

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
austinzh: Also we needs to test scalar. 
Tensorflow is capable to return scalar tensor shape of [] size of 1 as well.

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
austinzh: Since Tensorflow SparseTensor is not a child of Tensor,  Let me check what we can make the most use of it.

mleap-tensor/src/main/scala/ml/combust/mleap/tensor/Tensor.scala
ancasarb: thanks @voganrc, i am taking a look at this now.

ancasarb (commented):
hey @austinzh, thanks for the PR, i've added a couple of questions/comments regarding the tensorflow changes, while i finish reviewing the tensor related potential bug/issues. 

one more question, this upgrade to tf 2.0 is a breaking one? that's not an issue, just to know to mention it in the relevant section in the release notes.

thanks again!

mleap-tensorflow/src/main/scala/ml/combust/mleap/tensorflow/TensorflowModel.scala
```diff
@@ -21,12 +22,12 @@ case class TensorflowModel(@transient var graph: Option[tensorflow.Graph] = None
                            graphBytes: Array[Byte]) extends Model with AutoCloseable {
 
   def apply(values: Tensor[_] *): Seq[Any] = {
-    val garbage: mutable.ArrayBuilder[tensorflow.Tensor[_]] = mutable.ArrayBuilder.make[tensorflow.Tensor[_]]()
+    val garbage: mutable.ArrayBuilder[tensorflow.Tensor] = mutable.ArrayBuilder.make[tensorflow.Tensor]()
```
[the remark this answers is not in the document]
austinzh: Do you mean keep using the `dataType` in `MleapConverter.convert` method?
But we can get the dataType information form `tensor` object, use `schema` to cast it might cause trouble.

austinzh:
> hey @austinzh, thanks for the PR, i've added a couple of questions/comments regarding the tensorflow changes, while i finish reviewing the tensor related potential bug/issues.
> 
> one more question, this upgrade to tf 2.0 is a breaking one? that's not an issue, just to know to mention it in the relevant section in the release notes.
> 
> thanks again!

Hi @ancasarb, thanks for the comments.
I pass the test for the `example tensorflow wine quality model` which the pb file is generated by TF1.X.
So I think we this is not a breaking change. Model trained with TF1.X can still run.

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
austinzh: Added Scalar support in commit e33779ef0aeff76eb92fed83fc7e028c7d5ce5ca

mleap-tensorflow/src/test/scala/ml/combust/mleap/tensorflow/converter/MleapConverterSpec.scala
austinzh: I open and issue to track Sparse Tensor.
https://github.com/combust/mleap/issues/764

mleap-tensorflow/src/main/scala/ml/combust/mleap/tensorflow/converter/MleapConverter.scala
```diff
@@ -1,38 +1,100 @@
...
+          )
+        )
+      case Tensor.DoubleClass =>
+        TFloat64.tensorOf(
+          shape,
+          DataBuffers.of(
+              dense.values.asInstanceOf[Array[Double]], true, false
+          )
+        )
+      case Tensor.StringClass =>
```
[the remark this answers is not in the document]
ancasarb: is this still pending?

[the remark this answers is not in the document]
austinzh: @voganrc 
The version we use is not `tensorflow/tensorflow` is `tensorflow/java`, It has encode method when we call `TString.tensorOf(ndString)`

https://github.com/tensorflow/java/blob/50f7efa5821877668c2645886b0e4addef7306b6/tensorflow-core/tensorflow-core-api/src/main/java/org/tensorflow/types/TString.java#L82

mleap-tensorflow/src/main/scala/ml/combust/mleap/tensorflow/TensorflowModel.scala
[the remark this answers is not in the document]
voganrc: @ancasarb Are you okay with just inferring the type from the tensor that is passed in?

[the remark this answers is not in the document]
ancasarb: yeah, okay, let's see how this goes :) we can always add it back in i guess.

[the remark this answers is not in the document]
austinzh: Well `Casting.cast` is a good point.  I don’t have a very good solution for that. 
In the old code `dense.values.asInstanceOf[Array[Float]]`. I think we can not cast a buffer of type A to  a buffer B.
Maybe float32 --> int32 can, but not all it will cause segment fault. If we are going to support cast, we might needs to check source type and target type first, if they are the same, we copy the buffer, if they are not the same, well, it’s going to be slow, because we needs to assign one by one.
That’s the main reason I don’t want to do this in Tensor conversion, If they are not the same (which should be), I rather fail it

mleap-tensor/src/main/scala/ml/combust/mleap/tensor/Tensor.scala
austinzh: Hi @ancasarb I create a branch from master in my forked repo.
[test_nd_tensor](https://github.com/austinzh/mleap/tree/test_nd_tensor
)
Feel free to check it out and run `sbt mleap-tensor/test`. 
Here I attached a small piece of the logs.
[mleap-tensor-test.log](https://github.com/combust/mleap/files/6901973/mleap-tensor-test.log)

PR #762, opened 2021-06-15
Improved CountVectorizer Serialization Error

voganrc:
When serializing a `CountVectorizerModel` that has been fit on a `null` vocabulary, I currently get the following error:

https://pastebin.com/Y2Hc1kud

It's cryptic to figure out that `java.lang.IllegalArgumentException: requirement failed` in `Predef.scala` means that you cannot serialize a `StringList` containing `null`, so I improved the error messaging surrounding that


ancasarb (approved):
looks good, thank you!

Issue #751, opened 2021-04-21
Spark 3.1.1, scala 2.12 & xgboost 1.3.1 migration


emitc2h:
Hello mleap developers,

I've been busy migrating [transmogrif.ai](https://transmogrif.ai/) to use Spark 3.1.1, scala 2.12 and xgboost 1.3.1, and due to our heavy reliance on mleap I've had to play with the mleap master branch in order to make it work. I thought it would be a good time to share my findings cause we'll ultimately need a new mleap build in Maven to complete the migration.

What I've done so far is:
 * take the existing `spark-3.0.0` branch and rebase it on top of master
 * updated the spark (3.1.1), scala (2.12.10), akka (2.6.14 and 10.2.4 for akka-http) and xgboost (1.3.1) versions
 * made a few modifications to the code to adapt to spark & xgboost API changes
 
All of which was enough to fix all of our JSON serialization/deserialization issues although we still have a number of unit tests to fix and we may discover more issues related to mleap. What I got so far is for sure not sufficient to release a new build but I'm wondering if/when you could help me make it happen. I'm happy to submit either a new PR, push a new branch or update the `spark-3.0.0` branch with my changes. Let me know which you'd prefer.

jsleight:
I'll volunteer to help review what I'm sure will be a fun PR.  Given that the `spark-3.0.0` branch is now quite a bit out of sync with master, maybe easier to just make a new PR?

indranilr:
@emitc2h   could you please let me know if the fix is being pursued ? We have similar dependency on our product on MLEAP and are unable to upgrade Spark due to MLEAP dependency ?

anu-srivastava:
@emitc2h @jsleight I was wondering is there a timeline by when we can provide Spark3 support?

mengban:
> @emitc2h @jsleight I was wondering is there a timeline by when we can provide Spark3 support?

It seems that spark3 works well with `mlep 0.17.0`
here are my envs:
`Spark 3.1.2`
`XGBoost 1.4.1`
`mleap 0.17.0`
I trained a xgb classification model and export it by mleap successfully, however it doesn't work when I use xgb model as a stage in `Pipeline`.

indranilr:
> 
> 
> > @emitc2h @jsleight I was wondering is there a timeline by when we can provide Spark3 support?
> 
> It seems that spark3 works well with `mlep 0.17.0`
> here are my envs:
> `Spark 3.1.2`
> `XGBoost 1.4.1`
[...]

PR #765, opened 2021-07-14
Upgrade to spark 3.

### LINKED ISSUE #727: Spark 3.0
Hello, 

I see you have a spark3.0 branch in the works, but it hasn't been looked at in a while. Do you have a time estimate of it's release, or an estimate of the amount of work needed to get it ready?

jsleight:
#753 seems to not have much action, and was having issues with the xgboost upgrade.  So this PR leaves xgboost at it's current version, and focuses on Spark.  We can do xgboost as a separate PR.

Fixes #727 and sort of #751 

Some notes:

- Spark v3 added multicolumn input/output support for `Binarizer`, `StringIndexer`, and `StopWordsRemover`.  I did a "partial support" for this where we are now serializing any of the multi-column parameters like the thresholds/labels, but if we try to load the multicolumn into mleap, then an exception is thrown.
    - I /think/ it might need some tweaks to the spark deserialization for inputCols and outputCols.  I need to increase the test cases in a few places still but figured I'd post the PR to get any feedback.
    - Getting the full multicolumn support into mleap seems like it will be a larger change, so put that as out of scope for now.

- There are a few new Estimators/transformers in spark 3 ml, didn't add those in since we can always do it later.
- Tried to fix most of the deprecation warnings.


jsleight:
cc @ancasarb @voganrc @emitc2h

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/StringIndexerOp.scala
```diff
@@ -17,20 +17,36 @@ class StringIndexerOp extends SimpleSparkOp[StringIndexerModel] {
...
+      var result = model.
+        withValue("labels_length", Value.int(obj.labelsArray.length)).
         withValue("handle_invalid", Value.string(obj.getHandleInvalid))
+      obj.labelsArray.indices.foreach(
+        i => result = result.withValue(s"labels_array_$i", Value.stringList(obj.labelsArray(i)))
+      )
+      result
     }
 
     override def load(model: Model)
```
voganrc, line 29: Does the new deserialization logic get tested anywhere?
Do we have a good way to ensure backwards compatibility?

mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/BinarizerOp.scala
```diff
@@ -27,8 +27,14 @@ class BinarizerOp extends MleapOp[Binarizer, BinarizerModel] {
 
     override def load(model: Model)
                      (implicit context: BundleContext[MleapContext]): BinarizerModel = {
-      BinarizerModel(model.value("threshold").getDouble,
-        model.value("input_shapes").getDataShape)
+      val threshold: Option[Double] = model.getValue("threshold").map(_.getDouble)
+      val thresholds: Option[Seq[Double]] = model.getValue("thresholds").map(_.getDoubleList)
+      (threshold, thresholds) match {
```
voganrc, line 32: We should add tests for this

mleap-core/src/main/scala/ml/combust/mleap/core/ann/LossFunctions.scala
```diff
@@ -45,7 +45,7 @@ class SigmoidLayerModelWithSquaredError
   extends FunctionalLayerModel(new FunctionalLayer(new SigmoidFunction)) with LossFunction {
   override def loss(output: BDM[Double], target: BDM[Double], delta: BDM[Double]): Double = {
     ApplyInPlace(output, target, delta, (o: Double, t: Double) => o - t)
-    val error = Bsum(delta :* delta) / 2 / output.cols
+    val error = Bsum(delta *:* delta) / 2 / output.cols
```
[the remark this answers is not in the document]
jsleight: This is just `breeze`'s new api for the same function.  Mleap is actually getting the breeze dependency from spark, which upgraded to breeze 1.0.  So this change is just to fix the compile errors.

Really though, breeze should be an explicit dependency of mleap -- so I'll add that in next commit.

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/StringIndexerOp.scala
jsleight: The new deserialization logic does get tested by the spark parity tests, though it would only have a labelsArray of length one atm.  I'm working on adding a test with a multi-column input/output for the affected transformers, which would test a longer labelsArray.

Testing that the backwards compatibility works is trickier.  I think the only way would be to serialize an artifact from a spark 2 environment, and then check it into the mleap repo?  @ancasarb thoughts?

jsleight:
Ok, @voganrc @ancasarb I think we're ready for review.  Travis is passing 🎉 

I think the most unanswered question is what we want to do with testing the "legacy" deserialization behaviors for the MultiInOutFormat classes mentioned in the description.  E.g., if someone tries to deserialize a spark v2 StringIndexer object into a spark v3 environment, then it should be backwards compatible -- however that logic is never tested because we can't make a spark v2 StringIndexer object in the test suite.  There are two options here:
1. Just don't worry about it.  Look at the code and convince yourself it is correct.
2. We serialize a spark v2 StringIndexer bundle (e.g., by using mleap master branch) and check it into the mleap repo so we can use it tests.

I lean towards option 1 because I think the code is pretty simple and it avoids polluting the repo with these legacy test artifacts.

voganrc (approved):
LGTM.

I personally think it's fine to forgo committing legacy test artifacts.
`SparkParityBase` gives me confidence that the single input col / single output col deserialization pathway works

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/BinarizerOp.scala
```diff
@@ -22,25 +24,36 @@ class BinarizerOp extends SimpleSparkOp[Binarizer] {
       assert(context.context.dataset.isDefined, BundleHelper.sampleDataframeMessage(klazz))
 
       val dataset = context.context.dataset.get
-      model.withValue("threshold", Value.double(obj.getThreshold)).
-        withValue("input_shapes", Value.dataShape(sparkToMleapDataShape(dataset.schema(obj.getInputCol), dataset)))
+      var result = {
+        ParamValidators.checkSingleVsMultiColumnParams(obj, Seq(obj.inputCol), Seq(obj.inputCols))
+        if(obj.isSet(obj.inputCols)) {
+          model.withValue("input_shapes", Value.dataShape(ListShape()))
```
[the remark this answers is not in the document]
jsleight: This is the only one I encountered at least

.travis.yml
```diff
@@ -15,9 +15,9 @@ services:
 
 language: scala
 scala:
-  - 2.11.12
+  - 2.12.13
 jdk:
-  - oraclejdk8  # sbt +compile fails on jdk11 (xenial's default)
+  - openjdk8
```
ancasarb, line 20: should we do the java 11 upgrade before the release or 0.18? what do you think? @voganrc @jsleight

mleap-core/src/main/scala/ml/combust/mleap/core/feature/HashingTermFrequencyModel.scala
```diff
@@ -24,11 +24,17 @@ object HashingTermFrequencyModel {
...
+          hashUnsafeBytes(utf8, Platform.BYTE_ARRAY_OFFSET, utf8.length, seed)
+        } else {
+          hashUnsafeBytes2(utf8, Platform.BYTE_ARRAY_OFFSET, utf8.length, seed)
+        }
       case _ => throw new IllegalStateException("HashingTF with murmur3 algorithm does not " +
         s"support type ${term.getClass.getCanonicalName} of input data.")
     }
   }
+  val VERSION_1: Int = 1
+  val VERSION_2: Int = 2
```
ancasarb, line 37: maybe rename this to `HASH_UNSAFE_BYTES_VERSION_1` and `HASH_UNSAFE_BYTES_VERSION_2` to give a bit more clarity what the version applies to?

mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/HashingTermFrequencyOp.scala
```diff
@@ -20,14 +20,17 @@ class HashingTermFrequencyOp extends MleapOp[HashingTermFrequency, HashingTermFr
     override def store(model: Model, obj: HashingTermFrequencyModel)
                       (implicit context: BundleContext[MleapContext]): Model = {
       model.withValue("num_features", Value.long(obj.numFeatures))
-        .withValue("binary", Value.boolean(obj.binary))
+        .withValue("binary", Value.boolean(obj.binary)).
+        withValue("version", Value.int(obj.version))
```
ancasarb, line 24: do you think it's worth renaming to `hashUnsafeBytesVersion` so that it's not confused with the version of the transformer?

mleap-runtime/src/main/scala/ml/combust/mleap/runtime/transformer/Pipeline.scala
```diff
@@ -2,15 +2,15 @@ package ml.combust.mleap.runtime.transformer
 
 import ml.combust.mleap.core.Model
 import ml.combust.mleap.core.types.{DataType, NodeShape, StructField, StructType}
-import ml.combust.mleap.runtime.frame.{FrameBuilder, Transformer}
+import ml.combust.mleap.runtime.frame.{FrameBuilder, Transformer => FrameTransformer}
```
ancasarb, line 5: is the alias needed?

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/StopWordsRemoverOp.scala
```diff
@@ -2,42 +2,35 @@ package org.apache.spark.ml.bundle.ops.feature
...
-    new StopWordsRemover(uid = uid).setStopWords(model.getStopWords).setCaseSensitive(model.getCaseSensitive)
-  }
-
-  override def sparkInputs(obj: StopWordsRemover): Seq[ParamSpec] = {
-    Seq("input" -> obj.inputCol)
-  }
-
-  override def sparkOutputs(obj: StopWordsRemover): Seq[SimpleParamSpec] = {
-    Seq("output" -> obj.outputCol)
+    model
```
ancasarb, line 34: should we still be doing ` new StopWordsRemover(uid = uid).setStopWords(model.getStopWords).setCaseSensitive(model.getCaseSensitive)`?

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/BinarizerOp.scala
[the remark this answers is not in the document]
ancasarb: there are a few other transformers that use `input_shapes, for e.g. `VectorAssembler` etc.

can we do `model.withValue("input_shapes", Value.dataShape(sparkToMleapDataShape(dataset.schema(obj.getInputCols), dataset)))` and let sparkToMleapDataShape to derive the right shape?

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/BinarizerOp.scala
```diff
@@ -22,25 +24,36 @@ class BinarizerOp extends SimpleSparkOp[Binarizer] {
...
-    new Binarizer(uid = uid).setThreshold(model.getThreshold)
-  }
-
-  override def sparkInputs(obj: Binarizer): Seq[ParamSpec] = {
-    Seq("input" -> obj.inputCol)
-  }
-
-  override def sparkOutputs(obj: Binarizer): Seq[SimpleParamSpec] = {
-    Seq("output" -> obj.outputCol)
+    model
```
ancasarb, line 57: same as above, should it do `new Binarizer(uid = uid).setThreshold(model.getThreshold)`?

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/StringIndexerOp.scala
```diff
@@ -9,35 +9,45 @@ import org.apache.spark.ml.feature.StringIndexerModel
...
-      labels = model.labels).setHandleInvalid(model.getHandleInvalid)
-  }
-
-  override def sparkInputs(obj: StringIndexerModel): Seq[ParamSpec] = {
-    Seq("input" -> obj.inputCol)
-  }
-
-  override def sparkOutputs(obj: StringIndexerModel): Seq[SimpleParamSpec] = {
-    Seq("output" -> obj.outputCol)
+    model
```
ancasarb, line 51: and here as well.

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/BinarizerOp.scala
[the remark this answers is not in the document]
jsleight: iirc. sparkToMleapDataShape didn't work with the complex type so I had to special case things out.  But I'll take another look.

.travis.yml
jsleight: imo we should do the 0.18 release before java 11 for a couple main reasons:
1. It gets the spark 3 release out sooner, which many people have been requesting
2. It gives people an intermediate release which they can use on spark 3 and java 8.

ancasarb: 👍

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/StopWordsRemoverOp.scala
jsleight: (same idea for all 3 of these)

I had something like that at first, but this is kind of tricky and is incorrect/unnecessary.  How the code flow works.
1. We create the `StopWordsRemover` object in the `OpModel.load` method above this.  As part of this, the new `loadMultiInOutFormat` method sets the input/output parameters on that transformer object.
2. The `SimpleSparkOp.load` method calls `sparkLoad` with model equal to the output transformer object we created in step 1.  We do have the option of creating a new transformer object here, as long as we also would set the input/output parameters.  But this seems unnecessary and would result in having to duplicate all the multi in/out format multiplexing again?  So instead we just pass the model object though without changing it.
3.  The `SimpleSparkOp.load` method then calls `sparkInputs` and `sparkOutputs`.  For most transformers, `sparkInputs` and `sparkOutputs` are hardcoded since it just has to say what the transformers input/output parameters are called -- we can't hardcode because we need to support multi in/out formats.  The easiest way to handle this was to have the transformer object already set up with it's input/output parameters set up, which we did in step 1.
4. The `SimpleSparkOp.load` method then calls `SparkShapeLoader.loadShape` which has kind of a bad name and actually is responsible for setting the input/output parameters on the transformer.  For most transformers it would use the hardcoded parameter names provided by `sparkInputs` and `sparkOutputs`.  For the multi in/out format transformers it will just re-set the same parameter values we already set up.

So the key tricky thing is that for multi in/out formats we need to know the input/output parameters (step 3) before they are normally set (step 4).  Thus we need to set them up earlier outside of the usual code flow.  Step 1 was the logical choice since the serialization format can directly contain which format we're using.

I'm not really even sure why `sparkLoad` exists to be honest, but suspect there is some edge case transformer which needed to do something custom -- though am not sure why that custom logic couldn't be in step 1.  Or perhaps there is some historical reason that doesn't apply anymore?  In any case, most spark ops seemingly don't need the `sparkLoad` method at all and are creating a new transformer unnecessarily I think?

mleap-runtime/src/main/scala/ml/combust/mleap/runtime/transformer/Pipeline.scala
jsleight: I don't think it is strictly necessary, but there was a compiler warning about covering up names or something.  

there is technically a `ml.combust.mleap.runtime.Transformer` class, which is literally defined as `type Transformer = frame.Transformer`.  So it was covering itself.  Doesn't impact correctness, but I was trying to clean up warnings because they bother me 😄

mleap-spark/src/main/scala/org/apache/spark/ml/bundle/ops/feature/BinarizerOp.scala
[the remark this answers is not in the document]
jsleight: Right, looking again, the issue is that `sparkToMleapDataShape` expects a `spark.sql.types.StructField`  But we have multiple fields, so`dataset.schema(obj.getInputCols)` returns a `StructType` instead.  I thought about making `sparkToMleapDataShape` work with a StructType, but this ended up being rather complex and had a lot of downstream implications.

The way that VectorAssembler works is to create a DataShape for each of the inputcols, and then serialize a `dataShapeList` for the bundle.  This makes a lot of sense for VectorAssembler since it will always have an `inputCols` param (there is no option in spark for it to be single input format).  Doing that kind of approach might make more sense here too, though will mean we have to change the parameter name since the single vs multi format types will be different.  That seems acceptable though imo?

mleap-runtime/src/main/scala/ml/combust/mleap/runtime/transformer/Pipeline.scala
ancasarb: ah i see, okay :D

PR #774, opened 2021-09-14
Fix (List <-> Tensor) Casting When Base Types Match

voganrc:
I tried to use my previous casting PR and realized that:

```
Casting.cast(ListType(BasicType.Double), TensorType(BasicType.Double, Some(Seq())))
```

returns `None`. I also noticed that our casting tests lacked coverage for cases when `from.base == to.base`, so I added that in as well

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,6 +10,11 @@
 
 ### Other Changes
 
+# Release 0.18.1
+
+### Bug Fixes
+- Fix (List <-> Tensor) casting for identical base type
+
 # Release 0.18.0
 
 ### Breaking Changes
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,7 +13,7 @@
 # Release 0.18.1
 
 ### Bug Fixes
-- Fix (List <-> Tensor) casting for identical base type
+- Fix (List <-> Tensor) casting when base types match
 
 # Release 0.18.0
 
--- a/mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
+++ b/mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
@@ -1,6 +1,6 @@
 package ml.combust.mleap.core.types
 
-import ml.combust.mleap.tensor.{ByteString, Tensor}
+import ml.combust.mleap.tensor.{ByteString, Tensor, DenseTensor, SparseTensor}
 
 import scala.util.{Failure, Success, Try}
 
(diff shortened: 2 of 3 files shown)


mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
```diff
@@ -184,13 +184,41 @@ object Casting {
                   val cc = c.asInstanceOf[(Any) => ByteString]
                   (v: Any) => Tensor.denseVector(v.asInstanceOf[Seq[_]].map(cc).toArray)
               }
+            }
+          }
+        }.orElse {
```
voganrc, line 189: When the base cast returns `None`, we just want to reshape the list into a tensor without changing the types of each element

mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
```diff
@@ -184,13 +184,41 @@ object Casting {
...
+          Some {
+            Try {
+              from.base match {
+                case BasicType.Boolean =>
+                  (v: Any) => Tensor.denseVector(v.asInstanceOf[Seq[_]].map(_.asInstanceOf[Boolean]).toArray)
+                case BasicType.Byte =>
+                  (v: Any) => Tensor.denseVector(v.asInstanceOf[Seq[_]].map(_.asInstanceOf[Byte]).toArray)
```
voganrc, line 196: It's annoying to have to manually have specify these `asInstanceOf` calls when we don't actually want to cast the elements, but it is a consequence of making the function input `(v: Any)` and casting it to `Seq[_]`

If we don't manually re-cast each element, it will make a tensor of Java `Object` types, and our tests fail

mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
```diff
@@ -184,13 +184,41 @@ object Casting {
...
         }
       case (_: TensorType, _: ListType) =>
         baseCast(from.base, to.base).map {
-          _.map {
-            c => (v: Any) => v.asInstanceOf[Tensor[_]].rawValues.toList.map(c)
+          _.flatMap {
+            c => Try((v: Any) => v.asInstanceOf[Tensor[_]].rawValues.toList.map(c))
           }
+        }.orElse {
+          Some(Try((v: Any) => v.asInstanceOf[Tensor[_]].rawValues.toList))
```
voganrc, line 221: This works without manually specifying `_.asInstanceOf` like we did in the (List -> Tensor) direction

I believe this is because `Tensor` is not a native Scala class, so casting to one doesn't automatically cast all of the elements into Java `Object` types

mleap-core/src/test/scala/ml/combust/mleap/core/types/CastingSpec.scala
```diff
@@ -132,10 +141,10 @@ class CastingSpec extends FunSpec {
   for(((from, to, fromValue, expectedValue), i) <- castTests.zipWithIndex) {
     describe(s"cast from $from to $to - $i") {
       it("casts the scalar") {
-        val c = Casting.cast(ScalarType(from), ScalarType(to)).get.get
-        val oc = Casting.cast(ScalarType(from), ScalarType(to).nonNullable).get.get
-        val co = Casting.cast(ScalarType(from).nonNullable, ScalarType(to)).get.get
-        val oco = Casting.cast(ScalarType(from), ScalarType(to)).get.get
+        val c = Casting.cast(ScalarType(from), ScalarType(to)).getOrElse(Success((v: Any) => v)).get
```
voganrc, line 144: This just says that when `Casting.cast` returns `None` (indicating that no cast is necessary), then the tests should use the identity function instead

ancasarb (approved):

RELEASE_NOTES.md
```diff
@@ -10,6 +10,11 @@
 
 ### Other Changes
 
+# Release 0.18.1
```
ancasarb, line 13: i'll merge this when i do the new release.

mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
```diff
@@ -184,12 +184,58 @@ object Casting {
...
+                case BasicType.ByteString =>
+                  (v: Any) => Tensor.denseVector(v.asInstanceOf[Seq[_]].map(_.asInstanceOf[ByteString]).toArray)
+              }
+            }
           }
         }
       case (_: TensorType, _: ListType) =>
         baseCast(from.base, to.base).map {
-          _.map {
-            c => (v: Any) => v.asInstanceOf[Tensor[_]].rawValues.toList.map(c)
```
voganrc, line 192: @ancasarb It was too hard for me to figure out how to cast `SparseTensor` to `List`

`SparseTensor[String]` and `SparseTensor[_]` (which all of our casting functions technically accept) both impute missing values with `null`, while `SparseTensor[Double]` imputes missing values with 0

So it was too hard to figure out when we should fill null vs. when we should fill 0

Furthermore, once you have a List with nulls in it, you have to figure out how to cast each element to the target type, which adds another layer of complexity.

I ended up just raising an error when you try to cast a SparseTensor to a List, and we can revisit this problem later

PR #778, opened 2021-10-01
Upgrade to xgboost v1.3.1

jsleight:
Interesting things:

- v1.4.x didn't work for me and I couldn't figure out why.  I think it had issues somewhere in the JNI layer.  Thus this upgade only goes to the v1.3.x series.
- xgboost v1.3.3 is identical to v1.3.1 for jvm packages -- so this is the latest version of 1.3.x series for jvm packages.
- I had to reduce the floating point precision comparison in the tests.  I think 1e-6 is all we should normally expect though?


jsleight:
ugh, apparently the xgboost tests are flakey 😭 .  Was passing on my dev env, but failed in travis.  But by repeated runs in dev env, it sometimes passes and sometimes fails.

Let me dig in a bit.

PR #782, opened 2021-10-14
[Test Only] Databricks Runtime with Spark 3.2

wenfeiy-db:


wenfeiy-db:
Verify that it's probably the change `addSbtPlugin("com.frugalmechanic" % "fm-sbt-s3-resolver" % "0.11.0")` that causes the "2450.4 Python 3.6 tests" and "2450.5 Python 3.7 tests" to fail in CI

PR #789, opened 2021-11-19
Add Map support and MapEntrySelector 

jsleight:
Commit 1: adds map type support in all the relevant places
Commit 2: adds a MapEntrySelector to serve as a complete test that map support works correctly

### DIFF
--- a/mleap-core/src/main/scala/ml/combust/mleap/core/feature/MapEntrySelectorModel.scala
+++ b/mleap-core/src/main/scala/ml/combust/mleap/core/feature/MapEntrySelectorModel.scala
@@ -5,7 +5,7 @@ import ml.combust.mleap.core.reflection.MleapReflection.dataType
 import ml.combust.mleap.core.Model
 import ml.combust.mleap.core.types.{MapType, ScalarType, StructField, StructType}
 
-case class MapEntrySelectorModel[K: TypeTag, V: TypeTag](defaultValue: Any = None) extends Model {
+case class MapEntrySelectorModel[K: TypeTag, V: TypeTag](defaultValue: V = None) extends Model {
 
   def apply(m: Map[K, V], key: K): V = {
     m.getOrElse(key, defaultValue.asInstanceOf[V])
--- a/mleap-core/src/main/scala/ml/combust/mleap/core/feature/MapEntrySelectorModel.scala
+++ b/mleap-core/src/main/scala/ml/combust/mleap/core/feature/MapEntrySelectorModel.scala
@@ -1,0 +1,20 @@
+package ml.combust.mleap.core.feature
+
+import scala.reflect.runtime.universe.TypeTag
+import ml.combust.mleap.core.reflection.MleapReflection.dataType
+import ml.combust.mleap.core.Model
+import ml.combust.mleap.core.types.{MapType, ScalarType, StructField, StructType}
+
+case class MapEntrySelectorModel[K: TypeTag, V: TypeTag](defaultValue: Any = None) extends Model {
+
+  def apply(m: Map[K, V], key: K): V = {
+    m.getOrElse(key, defaultValue.asInstanceOf[V])
+  }
+
+  override def inputSchema: StructType = StructType(
+    "input" -> MapType(dataType[K].base, dataType[V].base),
+    "key" -> ScalarType(dataType[K].base).nonNullable
+  ).get
+
+  override def outputSchema: StructType = StructType(StructField("output", dataType[V])).get
+}
--- a/mleap-core/src/main/scala/ml/combust/mleap/core/types/DataType.scala
+++ b/mleap-core/src/main/scala/ml/combust/mleap/core/types/DataType.scala
@@ -131,3 +131,14 @@ case class ListType(override val base: BasicType,
 
   override def shape: ListShape = ListShape(isNullable)
 }
+
+
+case class MapType(val key: BasicType,
+                   override val base: BasicType,
+                   override val isNullable: Boolean = true) extends DataType {
+
+  override def setNullable(isNullable: Boolean): DataType = copy(isNullable = isNullable)
+  override def simpleString: String = "map"
+  override def printString: String = s"$simpleString(key=$key,base=$base,nullable=$isNullable)"
+  override def shape: ListShape = ListShape(isNullable)
+}
--- a/mleap-core/src/test/scala/ml/combust/mleap/core/reflection/MleapReflectionSpec.scala
+++ b/mleap-core/src/test/scala/ml/combust/mleap/core/reflection/MleapReflectionSpec.scala
@@ -29,6 +29,22 @@ class MleapReflectionSpec extends FunSpec {
       assert(dataType[java.lang.Long] == ScalarType.Long)
       assert(dataType[java.lang.Double] == ScalarType.Double)
       assert(dataType[Tensor[Double]] == TensorType(BasicType.Double))
+      assert(dataType[Map[String, String]] == MapType(BasicType.String, BasicType.String))
+      assert(dataType[Map[String, Int]] == MapType(BasicType.String, BasicType.Int))
+      assert(dataType[Map[String, Double]] == MapType(BasicType.String, BasicType.Double))
+      assert(dataType[Map[String, Boolean]] == MapType(BasicType.String, BasicType.Boolean))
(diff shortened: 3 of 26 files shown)


mleap-core/src/main/scala/ml/combust/mleap/core/types/DataType.scala
```diff
@@ -131,3 +131,14 @@ case class ListType(override val base: BasicType,
...
+
+
+case class MapType(val key: BasicType,
+                   override val base: BasicType,
+                   override val isNullable: Boolean = true) extends DataType {
+
+  override def setNullable(isNullable: Boolean): DataType = copy(isNullable = isNullable)
+  override def simpleString: String = "map"
+  override def printString: String = s"$simpleString(key=$key,base=$base,nullable=$isNullable)"
+  override def shape: ListShape = ListShape(isNullable)
```
jsleight, line 143: I used LIstShape here because I think the semantics are the same (i.e., it is a collection that can have a variable length).  Open to discussion though

mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/MapEntrySelectorOp.scala
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    override def load(model: Model)(implicit context: BundleContext[MleapContext]): MapEntrySelectorModel[_, _] = {
+      val keyBasicType = bundleToMleapBasicType(model.value("key_type").getBasicType)
+      val valueBasicType = model.value("value_type").getBasicType
+      val defaultValue = model.value("default_value").getAnyFromType(valueBasicType)
+      (keyBasicType, bundleToMleapBasicType(valueBasicType)) match {
+        case (BasicType.String, BasicType.Double) => MapEntrySelectorModel[String, Double](defaultValue)
+        case (k, v) => throw new UnsupportedOperationException(s"Can not load bundle of types $k, $v")
+      }
```
jsleight, line 35: So I tried really hard to make the MapEntrySelector be fully generic and dynamically compute the types.  I got everything to be generic except the Ops.  And I'm not sure if it is possible to make this part generic.  something something scala reflection something something compile time...

But clearly we can/should support more type settings, and I'd like to hear which ones we think are valuable.

jsleight: I had written some code to translate all the types back and forth between spark, mleap, and scala which all worked without issue.

The problem is that the types need to exist at runtime, which scala doesn't do usually (types are just for the compiler).  You can add `TypeTag` to things for the type to stick around, but that only seems to work for templated types.  This is why I could make all the transformer and model classes work since they have templated generics.  But I don't think we can make the ops be templatized (i.e., `MapEntrySelector[K,V]` is not a thing we can do).  Thus the code we have.

I did find some super gross options where you like exec a process to instantiate the types but I think just having the ops be case matched is better.

mleap-core/src/main/scala/ml/combust/mleap/core/types/DataType.scala
jsleight: We aren't serializing MapType anywhere.  The bundle just has to serialize the shape and node objects right?

mleap-core/src/main/scala/ml/combust/mleap/core/feature/MapEntrySelectorModel.scala
```diff
@@ -0,0 +1,20 @@
+package ml.combust.mleap.core.feature
+
+import scala.reflect.runtime.universe.TypeTag
+import ml.combust.mleap.core.reflection.MleapReflection.dataType
+import ml.combust.mleap.core.Model
+import ml.combust.mleap.core.types.{MapType, ScalarType, StructField, StructType}
+
+case class MapEntrySelectorModel[K: TypeTag, V: TypeTag](defaultValue: Any = None) extends Model {
```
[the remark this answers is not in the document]
jsleight: I don't remember.  I know I changed this back and forth a few times as I was dealing with generics and at various times it was necessary to have this be Any to make the compiler happy.  I'll check again.

mleap-core/src/main/scala/ml/combust/mleap/core/feature/MapEntrySelectorModel.scala
```diff
@@ -0,0 +1,20 @@
...
+
+  def apply(m: Map[K, V], key: K): V = {
```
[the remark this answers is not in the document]
jsleight: Correct, this is assuming you want to select a variable key per row.  If you just wanted a static key then you could do that prior to the mleap pipeline as part of the input preparation.

mleap-core/src/main/scala/ml/combust/mleap/core/feature/MapEntrySelectorModel.scala
```diff
@@ -0,0 +1,20 @@
...
+    m.getOrElse(key, defaultValue.asInstanceOf[V])
+  }
+
+  override def inputSchema: StructType = StructType(
+    "input" -> MapType(dataType[K].base, dataType[V].base),
```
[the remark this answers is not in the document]
jsleight: I had to learn too much about type reflection to write this PR....

mleap-core/src/test/scala/ml/combust/mleap/core/reflection/MleapReflectionSpec.scala
```diff
@@ -29,6 +29,22 @@ class MleapReflectionSpec extends FunSpec {
...
+      assert(dataType[Map[String, String]] == MapType(BasicType.String, BasicType.String))
+      assert(dataType[Map[String, Int]] == MapType(BasicType.String, BasicType.Int))
+      assert(dataType[Map[String, Double]] == MapType(BasicType.String, BasicType.Double))
+      assert(dataType[Map[String, Boolean]] == MapType(BasicType.String, BasicType.Boolean))
+      assert(dataType[Map[Int, String]] == MapType(BasicType.Int, BasicType.String))
+      assert(dataType[Map[Int, Int]] == MapType(BasicType.Int, BasicType.Int))
+      assert(dataType[Map[Int, Double]] == MapType(BasicType.Int, BasicType.Double))
+      assert(dataType[Map[Int, Boolean]] == MapType(BasicType.Int, BasicType.Boolean))
+      assert(dataType[Map[Double, String]] == MapType(BasicType.Double, BasicType.String))
+      assert(dataType[Map[Double, Int]] == MapType(BasicType.Double, BasicType.Int))
```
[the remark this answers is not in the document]
jsleight: Yes we are.  Also, Byte, Short, etc.
I can add them all if we want, but 🤷

mleap-core/src/main/scala/ml/combust/mleap/core/types/DataType.scala
jsleight: Had some offline discussions, and agreed we don't need a MapShape for now.

mleap-runtime/src/main/scala/ml/combust/mleap/bundle/ops/feature/MapEntrySelectorOp.scala
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    override def load(model: Model)(implicit context: BundleContext[MleapContext]): MapEntrySelectorModel[_, _] = {
+      val keyBasicType = bundleToMleapBasicType(model.value("key_type").getBasicType)
+      val valueBasicType = model.value("value_type").getBasicType
+      val defaultValue = model.value("default_value").getAnyFromType(valueBasicType)
+      (keyBasicType, bundleToMleapBasicType(valueBasicType)) match {
+        case (BasicType.String, BasicType.Double) => MapEntrySelectorModel[String, Double](defaultValue.asInstanceOf[Double])
+        case (BasicType.String, BasicType.Float) => MapEntrySelectorModel[String, Float](defaultValue.asInstanceOf[Float])
+        case (BasicType.String, BasicType.Long) => MapEntrySelectorModel[String, Long](defaultValue.asInstanceOf[Int])
```
voganrc, line 35: should be `asInstanceOf[Long]`

voganrc (approved):
LGTM except for one line!

PR #791, opened 2021-12-09
Implement Vector -> Tensor Casting

voganrc:
You can see [here](https://github.com/combust/mleap/blob/68cbf375968d9f55acb1d673a4c2390602c0274a/mleap-spark-base/src/main/scala/org/apache/spark/sql/mleap/TypeConverters.scala#L72) that MLeap's DataFrame -> LeapFrame conversion code defines `TensorType` as the MLeap equivalent of Spark's `VectorUDT`

You can see [here](https://github.com/combust/mleap/blob/68cbf375968d9f55acb1d673a4c2390602c0274a/mleap-spark-base/src/main/scala/org/apache/spark/sql/mleap/TypeConverters.scala#L31) though, that the `Vector` value for this field never actually gets coerced into a `Tensor[Double]`, and is just declared as such

This results in the following error getting thrown in the casting module:
```
java.lang.ClassCastException: org.apache.spark.ml.linalg.DenseVector cannot be cast to ml.combust.mleap.tensor.Tensor
```

I considered changing the `DataFrame -> LeapFrame` conversion logic at first, but decided against it because there are models like [VectorAssemblerModel](https://github.com/combust/mleap/blob/4175f631168118fab87d87f7a0e5a01cef6c6e5b/mleap-core/src/main/scala/ml/combust/mleap/core/feature/VectorAssemblerModel.scala#L90) that produce Vectors and assign a TensorType schema to them.

Therefore, I made the necessary changes in the Casting module itself, and you have to do a TensorType -> TensorType cast any time you want to convert a Vector into a Tensor

### DIFF
--- a/mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
+++ b/mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
@@ -1,6 +1,7 @@
 package ml.combust.mleap.core.types
 
 import ml.combust.mleap.tensor.{ByteString, Tensor, DenseTensor, SparseTensor}
+import org.apache.spark.ml.linalg.{DenseVector, SparseVector, Vector => SparkVector}
 
 import scala.util.{Failure, Success, Try}
 
@@ -81,6 +82,13 @@ object Casting {
     (BasicType.String, BasicType.Double) -> { (v: String) => if (v == "null" || v == "") null else v.toDouble }
   ).map {
     case (k, v) => (k, v.asInstanceOf[(Any) => Any])
+  }
+
+  def sparkVectorToMLeapTensor(sparkVector: SparkVector): Tensor[Double] = {
+    sparkVector match {
+      case v: DenseVector => DenseTensor[Double](v.values, Seq(v.size))
+      case v: SparseVector => SparseTensor[Double](v.indices.map(i => Seq(i)), v.values, Seq(v.size))
+    }
   }
 
   def baseCast(from: BasicType, to: BasicType): Option[Try[(Any) => Any]] = {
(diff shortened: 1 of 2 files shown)


mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
```diff
@@ -245,32 +253,88 @@ object Casting {
...
+                  }
+              }
+          }
+        }.orElse {
+          Some {
+            Try {
+              (v: Any) => {
+                v match {
+                  case _: SparkVector => sparkVectorToMLeapTensor(v.asInstanceOf[SparkVector])
+                  case _ => v
```
voganrc, line 334: This is a bit of a deviation from all the other casting logic.

Normally this module returns `None` when `from.base == to.base`, but in this case we must return a function that can cast a Vector into a Tensor

mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
```diff
@@ -245,32 +253,88 @@ object Casting {
...
+                    }
+                  }
                 case BasicType.String =>
                   val cc = c.asInstanceOf[(Any) => String]
-                  (v: Any) => v.asInstanceOf[Tensor[_]].mapValues(cc)
+                  (v: Any) => {
+                    v match {
+                      case _: SparkVector => sparkVectorToMLeapTensor(v.asInstanceOf[SparkVector]).mapValues(cc)
+                      case _ => v.asInstanceOf[Tensor[_]].mapValues(cc)
+                    }
```
jsleight, line 316: this code is copy-pasted quite a bit, could make it a helper?

jsleight (approved):

mleap-core/src/main/scala/ml/combust/mleap/core/types/Casting.scala
voganrc: I couldn't get the generic function definition to behave, so I'm just going to leave it as-is for now

PR #793, opened 2021-12-13
BugFix: StringIndexerModel Cache labels.length to avoid travel it everytime

austinzh:
[StringIndexerModel](https://github.com/combust/mleap/blob/4175f631168118fab87d87f7a0e5a01cef6c6e5b/mleap-core/src/main/scala/ml/combust/mleap/core/feature/StringIndexerModel.scala)
line 27 and 37 both call labels.length and labels is Seq[String] and implemented with List
So when input contains missing label, the complexity is O(N) very time.
We should change to use cache the val invalidValue = labels.length.

And also StringIndexer call stringToIndex.contains and stringToIndex.apply every time.
This will performance two lookup. It's totally unnecessary, because 99% percent contains return true
We can call stringToIndex.get then do a match on it's a Some or None.

### DIFF
--- a/mleap-core/src/main/scala/ml/combust/mleap/core/feature/StringIndexerModel.scala
+++ b/mleap-core/src/main/scala/ml/combust/mleap/core/feature/StringIndexerModel.scala
@@ -16,7 +16,7 @@ case class StringIndexerModel(labels: Seq[String],
                               handleInvalid: HandleInvalid = HandleInvalid.Error) extends Model {
   val stringToIndex: Map[String, Int] = labels.zipWithIndex.toMap
   private val keepInvalid = handleInvalid == HandleInvalid.Keep
-
+  private val invalidValue = labels.length
   /** Convert a string into its integer representation.
     *
     * @param value label to index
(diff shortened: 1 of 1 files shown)


mleap-core/src/main/scala/ml/combust/mleap/core/feature/StringIndexerModel.scala
```diff
@@ -16,28 +16,29 @@ case class StringIndexerModel(labels: Seq[String],
...
+  private val invalidValue = labels.length
   /** Convert a string into its integer representation.
     *
     * @param value label to index
     * @return index of label
     */
   def apply(value: Any): Int = if (value == null) {
     if (keepInvalid) {
-      labels.length
+      invalidValue
```
voganrc, line 27: Neat! I didn't realize measuring a List's length was O(N) in Scala:
https://stackoverflow.com/questions/8197655/why-doesnt-the-scala-list-have-a-size-field

mleap-core/src/main/scala/ml/combust/mleap/core/feature/StringIndexerModel.scala
```diff
@@ -16,28 +16,29 @@ case class StringIndexerModel(labels: Seq[String],
...
     } else {
       throw new NullPointerException("StringIndexer encountered NULL value. " +
         s"To handle NULLS, set handleInvalid to ${HandleInvalid.Keep.asParamString}")
     }
   } else {
     val label = value.toString
-    if (stringToIndex.contains(label)) {
```
voganrc, line 34: If `contains` and `apply` are both O(1), does it really matter that we do both every time?

voganrc (approved):
LGTM!

mleap-core/src/main/scala/ml/combust/mleap/core/feature/StringIndexerModel.scala
austinzh: HashingMap can be O(1), But not always. 
And if it's TreeMap, then It's O(Log(N)).

https://alvinalexander.com/scala/understanding-performance-scala-collections-classes-methods-cookbook/

PR #794, opened 2021-12-15
Fixing conversion for Mleap vectors with non-increasing indices

colmosh:
### Description
Mleap allows Sparse Tensors to be created with sequences of non-increasing indices; not so in Spark. `mleapTensorToSparkVector` fails if the mleap Tensor was created with such a sequence of indices. The following change fixes this: the new construction used here does not require the indices to be increasing.


jsleight (commented):
this lgtm, but please add a test case to prevent a regression in the future

jsleight (approved):

voganrc (approved):

Issue #792, opened 2021-12-13
Log4j vulnerability in MLeap runtime?


cappaberra:
The log4j core code is built into the MLeap runtime.  What is the MLeap project's action to use a patched version of log4j?

The current workaround is to add a system property to the JVM's startup, `-Dlog4j2.formatMsgNoLookups=true`, via the `JAVA_OPTS` environment variable.  But if someone relied on this functionality, a patched version would be required.

Thanks!

jsleight:
Great question! 

So mleap doesn't depend on log4j directly, though does have log4j coming in as a transitive dependency from a few places that I could find:

- log4j v1.2.17 which comes in via Spark 3.0.2.  v1.2.17 is not vulnerable to CVE-2021-44228, though is vulnerable to other exploits.  Once Spark updates ([their WIP PR](https://github.com/apache/spark/pull/34877) ) then obviously mleap will want to update as well.  We'll have to see the extent to which Spark decides to backport the fix, and if they only release something on the current version (i.e., 3.2.x) then mleap's upgrade will take some time since we have other dependencies to work too.  But if they backport to a spark 3.0.x then we can upgrade easily.
- log4j-to-slf4j v2.10.0 which comes in from spring boot; however, [according to the Spring people this is not vulnerable](https://github.com/spring-projects/spring-boot/issues/28978#issuecomment-990814359) because it only uses log4j-api not log4j-core. (Which is what I also see in mleap via `sbt dependencyTree`

So I think mleap is not vulnerable to CVE-2021-44228.

cappaberra:
Thanks @jsleight for the input!

With the latest CVEs since 1 week ago, adding CVE-2021-45046 and CVE-2021-45105 to the list, is it a fairly trivial amount of effort to have the the Spring Boot version bumped to at least a 2.5.x version that's not past its end-of-life?

According to https://spring.io/blog/2021/12/10/log4j2-vulnerability-and-spring-boot:
> Our upcoming v2.5.8 & v2.6.2 releases (due Dec 23, 2021) will pick up Log4J v2.17.0, but since this is such a serious vulnerability you may want to override our dependency management and upgrade your Log4J2 dependency sooner.
There are some good manually-configured workarounds in place for getting log4j v2.17.0 into one's dependency tree before 12/23/21. Maybe this is relatively low effort to do for MLeap?

jsleight:
[...]

PR #797, opened 2021-12-22
Spring and junit upgrades

jsleight:
Upgrading spring for #792, which broke some tests and needed junit upgrade as well.

Followed https://www.baeldung.com/junit-5-migration for junit v5 upgrades.


ancasarb (approved):

cappaberra:
Can the `CHANGELOG.md` also get an update covering what was changed in this MR?

jsleight:
> Can the `CHANGELOG.md` also get an update covering what was changed in this MR?

Yep, the changelog will be udated prior to the v0.19 release

PR #822, opened 2022-08-09
Upgrade xgboost dependency to be 1.6.1 version

WeichenXu123:
Signed-off-by: Weichen Xu <weichen.xu@databricks.com>

Upgrade xgboost dependency to be 1.6.1 version

### DIFF
--- a/.bsp/sbt.json
+++ b/.bsp/sbt.json
@@ -1,0 +1,1 @@
+{"name":"sbt","version":"1.4.9","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["/Library/Java/JavaVirtualMachines/jdk-11.0.2.jdk/Contents/Home/bin/java","-Xms100m","-Xmx100m","-classpath","/Users/weichen.xu/Library/Application Support/JetBrains/IdeaIC2021.2/plugins/Scala/launcher/sbt-launch.jar","xsbt.boot.Boot","-bsp"]}
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -17,7 +17,7 @@ object Dependencies {
   lazy val slf4jVersion = "1.7.30"
   lazy val awsSdkVersion = "1.11.1033"
   val tensorflowJavaVersion = "0.4.0" // Match Tensorflow 2.7.0 https://github.com/tensorflow/java/#tensorflow-version-support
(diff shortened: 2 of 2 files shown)


mleap-xgboost-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostClassificationModelParitySpec.scala
```diff
@@ -33,6 +33,8 @@ class XGBoostClassificationModelParitySpec extends SparkParityBase {
   // These params are not needed for making predictions, so we don't serialize them
   override val unserializedParams = Set("labelCol", "evalMetric", "objective")
 
+  override val excludedColsForComparison = Array[String]("prediction")
```
WeichenXu123, line 36: Exclude this column when comparing result.
The reason is xgboost might generate some prediction result with equal probability ( 0.5/0.5 ) on each label, then the prediction label might be nondeterministic 0 or 1.
So I exclude the prediction column.
But, we still compare the "probability" column which can ensure the model inference parity.

WeichenXu123:
CC @jsleight Could you help review and merge if it is good ?

mleap-xgboost-spark/src/test/scala/ml/dmlc/xgboost4j/scala/spark/mleap/XGBoostClassificationModelParitySpec.scala
jsleight: If we set the random seed can we get things to be deterministic for the test?

WeichenXu123: Seemingly no. There're some other non-deterministic factors such as multiple spark tasks run in parallelism and causes different sum order when aggregation.

Assuming it works, I still don't recommend the approach because next time xgboost upgrading we might need to update test again.

jsleight: 😢 but makes sense

jsleight (approved):

PR #829, opened 2022-09-19
Added backward compatibility fix for XGBoost models

chaitanya-basava:
Hi,

We have noticed a few issues while trying to load XGBoost models on latest mleap version, these models were trained using a mleap version built on scala 2.11.
The issue is due to **tree_limit** variable which wasn't getting stored during model serialisation in older mleap version but is a variable that is expected to be present in bundle file in the latest version. So, in order to fix this behaviour have updated the code to either read from bundle file or fallback to default value in case of its absence instead of failure during deserialisation.


jsleight (approved):
lgtm, thanks for the fix!

Issue #840, opened 2022-12-28
MathBinary input validation


praj-0:
[MathBinary](https://github.com/combust/mleap/blob/master/python/mleap/pyspark/feature/math_binary.py) pyspark transformer accepts inputA, inputB, defaultA and defaultB arguments but these are not mutually exclusive. So when user provides both inputA and defaultA it still creates the transformer but then sometimes fails in execution at inference time because the inferred schema for mleap is wrong. Ideally these arguments should be mutually exclusive so that wrong configuration is not provided.

PR #841, opened 2022-12-28
Allow only one of input or default in MathBinary

praj-0:
- Input validation for MathBinary pyspark transformer to allow only one of input column or default value
- Issue:https://github.com/combust/mleap/issues/840
- Unit testing done


python/mleap/pyspark/feature/math_binary.py
```diff
@@ -89,6 +89,10 @@ def __init__(
         #
         # if operation is not None, we can proceed to instantiate the scala classes
         if operation:
+            if (inputA and defaultA is not None) or \
```
austinzh, line 92: It's better if we check the final product of params.
Because use can set params in `__init__`, `setParams`, and set it one by one.
And set it one by one is actually the most common use case.
```
binaryMathBinary().setOpertion(BinaryOperation.Multiply).setInputA("aa").setInputB("bb").setDefaultA(1).setDefaultB(2)
```

austinzh (approved):

PR #844, opened 2023-02-06
Build with openjdk 8

WeichenXu123:
Signed-off-by: Weichen Xu <weichen.xu@databricks.com>

Build with openjdk 8


WeichenXu123:
@jsleight 
Can we make mleap CI running against JDK 8 ? Databricks runtime is still using jdk 8 .

WeichenXu123:
Run `sbt test` with jdk 8 passed on my development ubuntu machine.

jsleight:
The latest tensorflow-java versions are only building for java 11 https://github.com/tensorflow/java/blob/master/README.md#tensorflowjava-version-support

So downgrading to java 8 would prohibit us from upgrading tensorflow.

WeichenXu123:
ok I see.

PR #847, opened 2023-02-23
Serialize objective for XGBoostClassificationModel

harupy:
Preserve `objective` when serializing `XGBoostClassificationModel`. `objective` is used in `transform`:

https://github.com/dmlc/xgboost/blob/a65ad0bd9c26ff0ce4cb842d3312064cd1e9eadc/jvm-packages/xgboost4j-spark/src/main/scala/ml/dmlc/xgboost4j/scala/spark/XGBoostClassifier.scala#L396

```
    if (getObjective.equals("multi:softmax")) {
```

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -17,7 +17,7 @@ object Dependencies {
   lazy val slf4jVersion = "1.7.36"
   lazy val awsSdkVersion = "1.11.1033"
   val tensorflowJavaVersion = "0.4.0" // Match Tensorflow 2.7.0 https://github.com/tensorflow/java/#tensorflow-version-support
-  val xgboostVersion = "1.6.1"
+  val xgboostVersion = "1.7.3"
   val breezeVersion = "1.2"
   val hadoopVersion = "2.7.4" // matches spark version
   val platforms = "windows-x86_64,linux-x86_64,macosx-x86_64"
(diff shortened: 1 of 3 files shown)


WeichenXu123 (approved):

WeichenXu123:
> - serializes/deserializes the Spark model properly *** FAILED ***
[info]   "[binary:logistic]" did not equal "[reg:squarederror]" xgbc_169002594330__objective is not equivalent (SparkParityBase.scala:210)

The objective param is not saved / loaded correctly.

harupy:
I think that's because of:

https://github.com/dmlc/xgboost/pull/7781/files#diff-d5d17d50ceefbbb4461619ee53227ddf6290b349bbbb3e99fbc42327292616cbL108

project/Dependencies.scala
```diff
@@ -17,7 +17,7 @@ object Dependencies {
   lazy val slf4jVersion = "1.7.36"
   lazy val awsSdkVersion = "1.11.1033"
   val tensorflowJavaVersion = "0.4.0" // Match Tensorflow 2.7.0 https://github.com/tensorflow/java/#tensorflow-version-support
-  val xgboostVersion = "1.6.1"
+  val xgboostVersion = "1.7.3"
```
harupy, line 20: Let's see what happens with the latest version

WeichenXu123:
CC @jsleight

WeichenXu123 (approved):

WeichenXu123 (approved):

project/Dependencies.scala
jsleight: nice!

jsleight (approved):

PR #866, opened 2023-08-30
Fix - Add zip slip validation

David-Fadida:
Add zip slip validation in `FileUtil.scala`

### DIFF
--- a/bundle-ml/src/main/scala/ml/combust/bundle/serializer/FileUtil.scala
+++ b/bundle-ml/src/main/scala/ml/combust/bundle/serializer/FileUtil.scala
@@ -6,7 +6,6 @@ import java.util.zip.{ZipInputStream, ZipOutputStream}
 /**
   * Created by hollinwilkins on 9/11/16.
   */
-@deprecated("Prefer ml.combust.bundle.util.FileUtil object.")
 case class FileUtil() {
   import ml.combust.bundle.util.{FileUtil => FileUtils}
   @deprecated("use FileUtil.rmRF(Path).")
--- a/bundle-ml/src/main/scala/ml/combust/bundle/serializer/FileUtil.scala
+++ b/bundle-ml/src/main/scala/ml/combust/bundle/serializer/FileUtil.scala
@@ -6,6 +6,7 @@ import java.util.zip.{ZipInputStream, ZipOutputStream}
 /**
   * Created by hollinwilkins on 9/11/16.
   */
+@deprecated("Prefer ml.combust.bundle.util.FileUtil object.")
 case class FileUtil() {
   import ml.combust.bundle.util.{FileUtil => FileUtils}
   @deprecated("use FileUtil.rmRF(Path).")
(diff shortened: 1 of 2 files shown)


bundle-ml/src/main/scala/ml/combust/bundle/serializer/FileUtil.scala
```diff
@@ -29,7 +29,11 @@ case class FileUtil() {
       if(entry.isDirectory) {
         new File(dest, entry.getName).mkdirs()
       } else {
+        val destCanonical = dest.getCanonicalPath()
         val filePath = new File(dest, entry.getName)
+        if(!filePath.getCanonicalPath().startsWith(destCanonical + File.separator)){
+          throw new Exception("Entry is outside of the target dir: " + e.getName())
```
jsleight, line 35: test suite showing this as not compiling.  Think it needs to be `entry.getName()`?

jsleight (changes requested):

bundle-ml/src/main/scala/ml/combust/bundle/serializer/FileUtil.scala
David-Fadida: Sure

jsleight:
@David-Fadida we just shipped #864 to ready the repo for scala 2.13, which results in a merge conflict.  Mind updating your PR?  Thanks!

David-Fadida:
@jsleight Can you review the changes?

jsleight (approved):

srmish-jfrog:
@jsleight  since we are a CNA, we would like to assign a CVE to this issue/fix, is that acceptable?

jsleight:
> @jsleight since we are a CNA, we would like to assign a CVE to this issue/fix, is that acceptable? (I'm from David's team)

Sounds good.

srmish-jfrog:
@jsleight Are you planning to publish a new release with this fix? I don't want to publish the CVE before there is a fixed version available

jsleight:
> @jsleight Are you planning to publish a new release with this fix? I don't want to publish the CVE before there is a fixed version available

Apologies, I'll try to get a new release out today.

jkylekelly:
The POC appears to be for the PyPi ecosystem, but the advisory is solely for Maven.  Is this a mistake?

jkylekelly:
@jsleight, I'm not familiar with this project, but this scala code is also used by the PyPi hosted dependency?

If so, I'd expect the affected ecosystems in the security advisory to include all ecosystems where the project's vulnerable code can be used.  This appears to be PyPi and Maven. 

The reason is that when a developer imports the PyPi package into their project, security tools will not flag it as vulnerable because it's not listed as an affected ecosystem.

I appreciate your response & any clarifications.

jsleight:
The pypi package can end up using the scala code, however users would need to explicitly install the scala package into their spark jvm.  The scala code is not bundled within the python package.  I /think/ you could only upgrade the scala package it would still be compatible with the old python package but not contain the CVE (don't remember 100% off hand if there were any changes to the python side in the v0.23.1 release, would need to consult the release notes)

So in a pure technical sense, the pypi artifact does not contain a CVE, though I can understand the perspective that it is strongly related -- and certainly we would recommend that people upgrade both the scala and python together.

I'll defer to @srmish-jfrog about how they usually tag CVE's for artifacts split across languages in this manner.

PR #860, opened 2023-07-10
[WIP] Scala 2.13 compatible build

db-scnakandala:

Issue #811, opened 2022-04-08
Will Mleap support Scala 2.13?


masi-ghodrati:
We have planned to use Mleap in our framework. Our framework uses Scala 2.13. We can't downgrade to 2.12, as we use some of the new features. I would like to know if you plan on upgrading to Scala 2.13 in the near future?
? 
And if there is any documentation/resource that can help us upgrade to the right Scala version and make Combust compatible with our framework?

jsleight:
We are on spark 3.2 which is cross compatible for scala 2.12 and 2.13.  If you (or anyone) wants to submit a PR to cross compile mleap for both scala 2.12 and 2.13, then I'm happy to review it and get it merged.

Skimming through the [scala migration guide](https://docs.scala-lang.org/overviews/core/collections-migration-213.html) it seems the biggest changes are w.r.t. some collections APIs that I suspect mleap doesn't use too much.

masi-ghodrati:
thank you for your prompt response.I'm gonna work on that.thanks again

masi-ghodrati:
I tried to upgrade mleap to 2.13,there are some dependancies which are not supported for scala 2.13 such as xgboost4j.It's a blocker for updating the whole modules to Scala 2.13. we just use runtime module in our framework,Do you think it's possible to update only this module to 2.13  and make it compatible with our framework?

masi-ghodrati:
There are some dependencies which are not supported by Scala 2.13, such as xgboost4j. It's a blocker for updating the entire modules to Scala 2.13. Since our framework only uses the runtime module, do you think we could update it to 2.13 and make it compatible with our framework?

jsleight:
Ah, that makes sense.  In principle I think you can make only some of the modules be cross-compiled with scala 2.13 if you tweaked https://github.com/combust/mleap/blob/master/project/MleapProject.scala.  Right nowthe build settings are all shared from `Common.buildSettings`, but if you want to make a new `Common.buildSettingsNoScala2_13` and use that for the xgboost modules then I think that is ok for now.

masi-ghodrati:
thank you

nicster:
It seems that xgboost4j now added Scala 2.13 support: https://github.com/dmlc/xgboost/pull/9099

dotbg:
This issue was one of the motivators to open that PR :) 
We can test things with the [snapshot](https://s3-us-west-2.amazonaws.com/xgboost-maven-repo/list.html?prefix=snapshot/) version.
```
[...]

PR #880, opened 2024-12-31
Scala 2.13: Second take

amivanoff:
Let's enter 2025 with Scala 2.13 support! 😊 Happy new year! :christmas_tree::fireworks::christmas_tree::tada::christmas_tree::gift::christmas_tree:

Second take on Scala 2.13, based on unfinished PR #860. Related to #811.

1. Some sort of hand-made cross-compilation inside:

Scala 2.13: compiles and passes all tests (as far as I understand).
 - But many warnings left in code to maintain compatibility with 2.12
 - And also it blows IntelliJ Idea's mind: it detects some strange errors

Scala 2.12: compiles and passes all tests (as far as I understand).
 - But in requires to comment `import scala.collection.parallel.CollectionConverters._` in [OneVsRest.scala](https://github.com/combust/mleap/compare/master...amivanoff:mleap:scala-2.13-2nd-take?expand=1#diff-3be34f4fe15c2806a337b26e71bdb427d7169287e761bfece3861ad4e0596c16).

2. Due to immutable Seq in transformers in 2.13, there was a problem with MinMaxScalerPipelineParitySpec test -- Array values in `ArrayRow`s should be immutable. I made some fix in [TypeConverters.scala](https://github.com/combust/mleap/compare/master...amivanoff:mleap:scala-2.13-2nd-take?expand=1#diff-26ecd6d303ee73f799b7bbd1c15167d7dc724e7ab4fd4bc93160dc826313a427) to ensure immutability. Not sure if it is the right place.

3. No automatic cross-compilation in Travis due to 2. Don't know what is the best practice to handle cross-compilation in the Scala world.

There are at least two variants:
- Conditional Compilation [Some hacks, no official way](https://github.com/scala/scala-dev/issues/640)
  - We need hardcore low level ifdefs "before parser" preprocessor which works with TEXT LINES inclusion/exclusion (imports lines, etc,). Not high-level "type annotations" and types inclusion/exclusion.
- Separate branches -- it seems it's an overkill
- Separate folders -- also too clunky

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ services:
 
 language: scala
 scala:
-  - 2.12.18
+  - 2.13.15
 jdk:
   - openjdk11
 
--- a/bundle-hdfs/src/main/scala/ml/bundle/hdfs/HadoopBundleFileSystem.scala
+++ b/bundle-hdfs/src/main/scala/ml/bundle/hdfs/HadoopBundleFileSystem.scala
@@ -30,7 +30,7 @@ object HadoopBundleFileSystem {
   }
 
   def createSchemes(config: Config): Seq[String] = if (config.hasPath("schemes")) {
-    config.getStringList("schemes").asScala
+    config.getStringList("schemes").asScala.toSeq
   } else { Seq("hdfs") }
 }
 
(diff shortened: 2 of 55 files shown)


zhenglaizhang:
@amivanoff Could you help share the plan for the PR? We are also looking forward to the Scala 2.13 support. Really appreciate for your effort on this!

amivanoff:
@zhenglaizhang , @Urvashi2303 Feel free to take the third take! Unfortunately, I have no plans to take it further.

